PR #20, opened 2022-09-30
test: add broker integration tests

bayological:
### Description
Add integration tests tying all McMint related work together.

Things to test:
- [x] Stable to Collateral swaps
- [x] Stable to Stable swaps
- [x] Bucket updates
- [x] Discovery functions 

### Other changes

N/A

### Tested

It is tests

### Related issues

- Fixes https://github.com/mento-protocol/mento-general/issues/43

### Backwards compatibility

N/A

### Documentation

N/A

### DIFF
--- a/test/integration/BrokerIntegration.t.sol
+++ b/test/integration/BrokerIntegration.t.sol
@@ -34,31 +34,7 @@ contract BrokerIntegrationTest is Test, McMintIntegration, TokenHelpers {
     deal(address(celoToken), address(reserve), 10**24); // Gift 1Mil to reserve
     deal(address(usdcToken), address(reserve), 10**24); // Gift 1Mil to reserve
   }
-}
 
-contract BrokerIntegrationTest_discovery is BrokerIntegrationTest {
-  function test_getExchangeProviders_shouldReturnProviderWithCorrectExchanges() public {
-    // Verify this function returns the BiPoolManager with valid poolExchanges as configured.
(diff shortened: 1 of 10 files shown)


test/integration/BrokerIntegration.t.sol
```diff
@@ -0,0 +1,292 @@
...
+      bucketIn = exchange.bucket0;
+      bucketOut = exchange.bucket1;
+    } else {
+      bucketIn = exchange.bucket1;
+      bucketOut = exchange.bucket0;
+    }
+
+    // Get expected out.. ¯\_(ツ)_/¯
+    // Maybe use broker.getAmountOut
+    expectedOut = exchange.pricingModule.getAmountOut(bucketIn, bucketOut, exchange.config.spread.unwrap(), amountIn);
```
bowd, line 67: Yeah I think using `broker.getAmountOut` here is fair.
I don't think we should make expectations on the amounts themselves in the test, that's part of the components. 
But in an integration test I think it's fair to test that `broker.getAmountOut` ends up being the `amountOut` of the immediate swap.

bowd (approved):
LGTM, I would indeed use `broker.getAmountOut` just to make it feel more consistent as an integration test, testing the boundary of the system.

But yeah a big bunch of the setup there is my own code so I don't think it makes sense for me to review it, so I'm hoping somebody else put some eyes on it.

philbow61 (commented):
Looks good to me👌🏻💪🏻

PR #27, opened 2022-10-18
test: migrate StableToken tests to foundry

### LINKED ISSUE #7: Migrate StableToken tests
**Description**
Convert the existing Stabletoken.ts test in the Celo monorepo to a solidity test using foundry

**Acceptance Criteria**
- mento-core repo contains Stabletoken tests that are all passing

ninocomputer:
### Description

Refactored StableToken tests from TS to Foundry

### Related issues

- Fixes #7


test/StableToken.t.sol
```diff
@@ -1,13 +1,634 @@
...
+    registry.setAddressFor("GrandaMento", grandaMento);
+  }
+
+  function mintAndAssert(address to, uint256 value) public {
+    changePrank(to);
+    testee.mint(to, value);
+    assertEq(testee.balanceOf(to), value);
+    assertEq(testee.totalSupply(), value);
+  }
+
```
philbow61, line 157: Enjoyed the following part with the mintAndAssert function😎

test/StableToken.t.sol
```diff
@@ -1,13 +1,634 @@
...
+    vm.roll(block.number + 1);
+    uint256 newBalance = testee.balanceOf(sender);
+    assertEq(newBalance, balance);
+  }
+
+  function test_balanceOf_withInflation_shouldFetchCorrectBalance() public {
+    changePrank(deployer);
+    testee.setInflationParameters(inflationRate, 1 weeks);
+    uint256 adjustedBalance = testee.balanceOf(sender);
+    assertEq(adjustedBalance, 1e20);
```
philbow61, line 335: shouldn't the balance change based on the set inflationRate? 
Not sure if I understood everything right 😌

test/StableToken.t.sol
```diff
@@ -1,13 +1,634 @@
...
+
+  function burnAndAssert(address to, uint256 value) public {
+    changePrank(to);
+    vm.expectEmit(true, true, true, true);
+    emit Transfer(to, address(0), testee.valueToUnits(value));
+    testee.burn(value);
+    assertEq(testee.balanceOf(to), value);
+    assertEq(testee.totalSupply(), value);
+  }
+
```
philbow61, line 384: Enjoyed the following part with the burnAndAssert function😎

test/StableToken.t.sol
```diff
@@ -1,13 +1,634 @@
...
+    changePrank(deployer);
+    setUpInflation(inflationRate);
+
+    changePrank(sender);
+
+    vm.expectEmit(true, true, true, true);
+    emit InflationFactorUpdated(8333333333333333333333333, 1 weeks + initTime);
+    testee.approve(receiver, transferAmount);
+  }
+
```
philbow61, line 492: I think here one test from the celo mono-repo is missing  
`allowance() should return allowance`

test/StableToken.t.sol
```diff
@@ -1,13 +1,634 @@
...
+  function test_balanceOf_withInflation_shouldFetchCorrectBalance() public {
+    changePrank(deployer);
+    testee.setInflationParameters(inflationRate, 1 weeks);
+    uint256 adjustedBalance = testee.balanceOf(sender);
+    assertEq(adjustedBalance, 1e20);
+  }
+}
+
```
philbow61, line 338: I am unsure about the following part. Don't really understand where these Results come from. The Original tests form the mono-repo seem different, but maybe I am missing something.

philbow61 (commented):
All in all looks good to me :) I left some comments where I felt unsure. Hope maybe @bayological or @bowd  can take a look at it as well

test/StableToken.t.sol
```diff
@@ -1,13 +1,634 @@
...
+    assertEq(rate, newInflationRate);
+    assertEq(updatePeriod, newUpdatePeriod);
+    assertEq(lastUpdated, initTime);
+  }
+
+  function test_setInflationParameters_shouldEmitEvent() public {
+    uint256 newUpdatePeriod = 1 weeks + 5;
+
+    changePrank(deployer);
+    vm.expectEmit(false, false, false, true);
```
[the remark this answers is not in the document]
ninocomputer: yeah, good catch!

test/StableToken.t.sol
ninocomputer: yes, you did understand it right! Time travel is necessary for the inflation factor to update.

test/StableToken.t.sol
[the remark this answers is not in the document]
bayological: @ninocomputer unresolved this, so it's not missed as the params have not been updated yet.

bayological:
> All in all looks good to me :) I left some comments where I felt unsure. Hope maybe @bayological or @bowd can take a look at it as well

I will look over the comments, but for this PR it makes sense for @bowd to review as I also did some of the tests.

test/StableToken.t.sol
ninocomputer: in monorepo in this particular test, I believe inflation factor is 1.005 * 10**24 but here the factor is 0.995 * 10**24 so the result is obviously different for this reason

bowd: This is because the mock you set up for `fractionMulExp` doesn't actually calculate the right values.
You can take a look [here](er/packages/vm/lib/evm/precompiles/fc-fractionmulexp.ts) for what the `fractionMulExp` should do. Essentially it's `a * b^exp` where a and b are both fractions, normalized to a number of decimals. So the parameters passed to the precompile are:
- aNumerator
- aDenominator
- bNumerator
- bDenominator
- exponent
- decimals

So the actual mock that you're setting now: 
```solidity
    ph.mockReturn(
      FRACTION_MUL,
      keccak256(abi.encodePacked(uint256(1e24), uint256(1e24), uint256(995e21), uint256(1e24), uint256(1), uint256(18))),
      abi.encode(uint256(100), uint256(12))
    );
``` 
Doesn't make sense for the values.

You have:
a = 1e24 / 1e24
b = 995e21 / 1e24
exp = 1
decimals = 18

Return value should be:
a * b^exp = a * b `(exp = 1)`
a * b = 995e21 / 1e24

And if we normalize for 18 decimals:

resultNumerator = 995e21 * 1e18 / 1e24 = 995e15
resultDenominator = 1e18

So the mock should be:
```solidity
    ph.mockReturn(
      FRACTION_MUL,
      keccak256(abi.encodePacked(uint256(1e24), uint256(1e24), uint256(995e21), uint256(1e24), uint256(1), uint256(18))),
      abi.encode(uint256(995e15), uint256(1e18))
    );
``` 

Then I would suggest replicating the other specs as well with the other inflation factors and time periods (which will result in higher exp values).

I would do this by: 
1.  writing a missing test case
2. running it, which will result in an error because there's no mock registered for that particular call of the precompile
3. look at the output of the failure to extract the arguments for the precompile
4. calculate the correct response using the formulas above
5. add a precompile mock
6. rinse and repeated until there are no more test cases

---

Long term this would be cleaner if the precompile test helpers actually implemented the a*b^exp calculation somehow and there are ways using `ffi` in forge to call to an external script, but that has more complexity and we're looking to move away from inflation anyway which might end up removing all of this code so I would keep it simple now.

bowd (changes requested):
Great job overall! 
There's that complexity with the `fractionMulExp` precompile that I would like us to cover. Let me know if the comment I left makes sense and you know how to approach it.

test/StableToken.t.sol
```diff
@@ -3,11 +3,637 @@ pragma solidity ^0.5.13;
...
+  address receiver;
+  string comment;
+  uint256 mintAmount = 100 * 10**18;
+
+  function setUp() public {
+    super.setUp();
+
+    sender = actor("sender");
+    receiver = actor("receiver");
+    comment = "pineapples absolutely do not belong on pizza";
```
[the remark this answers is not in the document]
ninocomputer: NEVER

test/StableToken.t.sol
ninocomputer: yes, yes now looking at the encoded values I realise that initially they were placeholders that I didn't calculate just to get the mock working and forgot to change. Now, this makes sense why the inflation factor was off.

bowd: @philbow61 can you pick this up and get this PR across the finish line, so that @ninocomputer can focus on the Circuit Breaker work?

philbow61: sure! I am on it

philbow61: @bowd should be all fixed now. I also added the code for making the broker capable of minting Stabletokens and the corresponding tests.

bowd: Awesome! Great job!

bowd (approved):

PR #45, opened 2022-10-26
chore: revert codebase to pre StableToken bucket cap

### LINKED ISSUE #43: Revert stable bucket cap size changes
**Description**
An [issue](https://github.com/celo-org/celo-monorepo/issues/5849) was created some time ago to add a cap to the stable bucket size, but never made its way to Mainnet. In order to maintain consistency this change should be reverted in the Celo mono repo and Mento core repos. The issue the change was supposed to be address will be handled by the new trading limits that will be brought in with McMint.

**Acceptance Criteria**

- [x] Revert the changes from the PR https://github.com/celo-org/celo-monorepo/pull/9318 in the celo-monorepo
- [x] Revert the changes for the same PR in mento-core (as a hotfix).
- [x] Remove references to code in tests
- [x] Remove references(if any) to code in contractKit

ninocomputer:
### Description

Due to the recent design decisions, revert `Exchange.sol`, contracts derived from it and exchange test to its state before cap for a  stableToken bucket was implemented. 

### Related issues

- Fixes #43

### Backwards compatibility

should be backwards compatible

### DIFF
--- a/test/Exchange.t.sol
+++ b/test/Exchange.t.sol
@@ -29,8 +29,6 @@ contract ExchangeTest is Test, WithRegistry, TokenHelpers {
     event SpreadSet(uint256 spread);
     event ReserveFractionSet(uint256 reserveFraction);
     event BucketsUpdated(uint256 celoBucket, uint256 stableBucket);
-    event StableBucketMaxFractionSet(uint256 stableBucketMaxFraction);
-    event MinSupplyForStableBucketCapSet(uint256 minSupplyForStableBucketCap);
 
     address deployer;
     address rando;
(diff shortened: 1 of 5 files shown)


test/Exchange.t.sol
```diff
@@ -29,8 +29,6 @@ contract ExchangeTest is Test, WithRegistry, TokenHelpers {
     event SpreadSet(uint256 spread);
```
ninocomputer, line 29: since there seems to be a problem in CI, it's failing

bowd (changes requested):
The changes look ok to me but we need to follow a slightly different merge strategy for this.
Because these changes are essentially needed on both the `main` and `develop` branches we need to think of this in terms of a hotfix. [This is a good example](https://www.theserverside.com/blog/Coffee-Talk-Java-News-Stories-and-Opinions/GitFlow-Hotfix-Branch-Example-Start-Finish) of a how to do a hotfix in the git flow methodology. Note that the example uses the actual `git-flow` plugin that we decided not to use for simplicity so you can ignore the actual commands, which are just light wrappers around 1 or more git operations. So essentially what I would do is:
- Create a branch `hotfix/revert-stable-token-cap` branched from `main` (important)
- Cherry-pick the relevant commit from this branch to the hotfix branch
- Open 2 PRs, one to `main` one to `develop`

@philbow61 can you also take this pls? ❤️

PR #50, opened 2022-10-27
test: migrate SortedOracles tests

philbow61:
### Description
- migrated the SortedOracles test from the celo mono-repo to foundry 

### Other changes
- fixed an off by one error in the Exchange.t.sol
- before the amount had a range from `[0 - sellerCeloBalance]` using Fuzz testing 
- in the case of the amount being equal to `sellerCeloBalance` the call isn't failing because the `'transfer value exceeded sender's allowance for recipient'` but because the `'transfer value exceeded balance of sender'`

### Related issues
- Fixes #[18] [Ticket](https://github.com/orgs/mento-protocol/projects/5/views/1)

### DIFF
--- a/contracts/BiPoolManager.sol
+++ b/contracts/BiPoolManager.sol
@@ -1,0 +1,475 @@
+pragma solidity ^0.5.13;
+pragma experimental ABIEncoderV2;
+
+import { Ownable } from "openzeppelin-solidity/contracts/ownership/Ownable.sol";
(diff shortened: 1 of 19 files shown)


bayological:
@philbow61 Looks like you branched from the McMint feature branch instead of develop 🙃.

bayological (changes requested):
@philbow61 Looks like you branched from the McMint feature branch instead of develop 🙃

PR #51, opened 2022-10-31
test: migrate sorted oracles tests

philbow61:
Description

migrated the SortedOracles test from the celo mono-repo to foundry
Other changes

fixed an off by one error in the Exchange.t.sol
before the amount had a range from [0 - sellerCeloBalance] using Fuzz testing
in the case of the amount being equal to sellerCeloBalance the call isn't failing because the 'transfer value exceeded sender's allowance for recipient' but because the 'transfer value exceeded balance of sender'
Related issues

Fixes #[18] [Ticket](https://github.com/orgs/mento-protocol/projects/5/views/1)

### DIFF
--- a/test/SortedOracles.t.sol
+++ b/test/SortedOracles.t.sol
@@ -1,0 +1,540 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.5.13;
+
+import { Test, console2 as console } from "celo-foundry/Test.sol";
+import { SortedOracles } from "contracts/SortedOracles.sol";
+
+import "../contracts/common/linkedlists/SortedLinkedListWithMedian.sol";
+import { FixidityLib } from "contracts/common/FixidityLib.sol";
+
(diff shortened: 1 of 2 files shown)


bayological (approved):
All good. Some minor comments on naming.

test/SortedOracles.t.sol
```diff
@@ -0,0 +1,540 @@
...
+        sortedOracles.removeOracle(token, oracle, 1);
+    }
+
+    function test_removeOracle_shouldRevertWhenAddressIsWrong() public {
+        submitNReports(1);
+        vm.expectRevert("token addr null or oracle addr null or index of token oracle not mapped to oracle addr");
+        sortedOracles.removeOracle(token, address(15), 0);
+    }
+
+    function test_removeOracle_shouldRevertWhenCalledByNonOwner() public {
```
philbow61, line 274:
```suggestion
    function test_removeOracle_whenCalledByNonOwner_shouldRevert() public {
```

PR #56, opened 2022-11-01
feat: McMint circuit breaker integration

### LINKED ISSUE #35: Circuit breaker integration with McMint
**Description**
The existing circuit breaker implementation is designed to work with the existing exchanges. With the McMint update, these exchanges will be deprecated, so the circuit breaker needs to be modified to work with the McMint broker instead.

Existing branch: https://github.com/celo-org/celo-monorepo/tree/bayological/onchain-circuit-breaker

**Acceptance Criteria**

- [x] Existing circuit breaker code is migrated to mento-core repo
- [x] Circuit breaker stops trading via mento when a shift gt a calculated amount in oracle reports is detected
- [x] Breaker automatically resets itself after it has been tripped based on it's configured params
- [x] Tests are added to verify functionality works as expected

ninocomputer:
### Description

- [x] moving initial circuit breaker implementation from [celo-monorepo](https://github.com/celo-org/celo-monorepo/pull/9677) to mento core. 
- [x] Add reference to the BreakerBox in the BiPoolManager
- [x] Query the breaker box by using the oracleReportTarget variable in the pair details
- [x] Write tests to verify everything works as expected

_circuit breaker integration with McMint design doc can be found [here](https://www.notion.so/e62010a140154271a48aba711a9f897a?v=35252ec445274fcf93043604e39f6733&p=3ffd6d250a074efb8b952714c90ec64f&pm=s)_

### Other changes

N/A

### Tested

- Unit tests
- Integration tests

### Related issues

- Fixes #35


contracts/BiPoolManager.sol
```diff
@@ -145,6 +153,19 @@ contract BiPoolManager is IExchangeProvider, IBiPoolManager, Initializable, Owna
...
   }
 
+ /**
+   * @notice Check trading mode for a particular oracle report target
+   * @param exchange The exchange i.e PoolExchange to use
+   */
+  function checkTradingMode(PoolExchange memory exchange)public view {
+    if (address(breakerBox) != address(0)) {
+      require(
+        breakerBox.getTradingMode(exchange.config.oracleReportTarget) == TRADING_MODE_BIDIRECTIONAL,
```
bowd, line 163: Ok so naming ways there have been a few ongoing conversations and mental renamings of things so I want to take this opportunity to propose some naming changes across the board.

Oracles work with `rates`, these `rates`  are reported to `rateFeeds` from which a median is constructed. 
When an oracle client reports a rate it passes in a `rateFeedID` which is the identifier of the `rateFeed` to which to add the rate to. 

> Currently what I now call `rateFeedID` has historically been the StableToken address and in the SortedOracles it's referenced as `token`, for example [here](https://github.com/mento-protocol/mento-core/blob/develop/contracts/SortedOracles.sol#L250-L252), `medianRate` takes `token` parameters but this should ideally be renamed to `rateFeedID` across the board - and I think we can do this here and be backward compatible.

Circuit breakers are applied to `rateFeeds`, thus they should also save breakers based on the `rateFeedID` and use this naming throughout.

The `BiPoolManager.Pool` struct provides a virtual AMM that follows a reference rate. It has two related configuration params:
- `uint256 referenceRateResetFrequency` - which determines how often to reset the AMM buckets to the reference rate
- `address oracleReportTarget` - which should be renamed to `referenceRateFeedID` and be the `rateFeedID` of the price that this pool follows (i.e. it's reference rate).

--- 

In this PR you've currently used the `referenceRate` naming in the circuit breaker, this is because I used it when explaining the circuit breaker integration logic from the perspective of the BiPoolManager, and in that context, the poolManager does indeed want to check if it's reference rate feed is still active or the circuit breaker has sprung. But inside the Circuit Breaker itself it doesn't make sense to use `referenceRate` but rather `rateFeed`.

Hope this makes sense and you can make the renames in this PR.

bowd (commented):
Initial thoughts on naming changes to make things consistent, will follow with more notes.

contracts/BreakerBox.sol
```diff
@@ -14,18 +13,18 @@ import { Initializable } from "./common/Initializable.sol";
  * @title   BreakerBox
  * @notice  The BreakerBox checks the criteria defined in separate breaker contracts
  *          to determine whether or not buying or selling should be allowed for a
- *          specified referenceRateIDs. The contract stores references to all breakers
- *          that hold criteria to be checked, referenceRateIDs that
+ *          specified rateFeedIDs. The contract stores rate feed IDs to all breakers
```
bowd, line 16: You renamed `references` => `rate feed IDs` where `reference` was not in relation to the `referenceRate` naming. Te sentence should still be: "the contract stores references".

contracts/SortedOracles.sol
```diff
@@ -16,103 +16,103 @@ import "./common/linkedlists/SortedLinkedListWithMedian.sol";
  * @title Maintains a sorted list of oracle exchange rates between CELO and other currencies.
```
bowd, line 16: Let's actually revert this commit for now as making all the changes necessary here actually breaks compatibility.

ninocomputer: reverted

contracts/BreakerBox.sol
ninocomputer: done!

PR #57, opened 2022-11-03
feat: make broker capable of minting and burning STs

### LINKED ISSUE #32: Make the Broker capable of minting StableTokens
### Description

The Broker Contract needs permission to mint/burn mento stables when executing swaps. In order to not break the patterns used in StableToken we will do this by: 
- Registering the Broker Contract in the Registry with identifier `keccak256(abi.encodePacked("Broker"))` (see GRANDA_MENTO_REGISTRY_ID in StableToken.sol)
- Add another condition in the mint/burn functions of stable token checking if the caller is the broker address from registry

### Acceptence criteria

- StableToken si modified to look-up the Broker in the registry and allow minting/burning.
- Integration tests no longer use the `exchangeRegistryId` variable in StableToken  (set to "Broker") to allow minting.

philbow61:
### Description

- added the `BROKER_REGISTRY_ID` to the StableToken contract 
- both `mint()` and `burn()` now also check whether the caller is the broker address from the registry
- thereby allowing a registered broker to mint and burn 

### Tested

- the integration test mentioned in the acceptance criteria has not been migrated yet and therefore needs to be updated later on

### Related issues

- Fixes #32

### DIFF
--- a/contracts/StableToken.sol
+++ b/contracts/StableToken.sol
@@ -42,6 +42,7 @@ contract StableToken is
     event TransferComment(string comment);
 
     bytes32 constant GRANDA_MENTO_REGISTRY_ID = keccak256(abi.encodePacked("GrandaMento"));
+    bytes32 constant BROKER_REGISTRY_ID = keccak256(abi.encodePacked("Broker"));
 
     string internal name_;
     string internal symbol_;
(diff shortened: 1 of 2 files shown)


philbow61:
> it looks good, add integration test in BrokerIntegration.t.sol and unit test in StableToken.t.sol to verify that it works as expected. Then it's good to go.

The migrated StableToken.t.sol is currently on develop. Therefore I would suggest writing the same test for [mint](https://github.com/mento-protocol/mento-core/blob/dedb150aed90d44a4239d5a6e558fe6836da276c/test/StableToken.t.sol#L183) and [burn](https://github.com/mento-protocol/mento-core/blob/dedb150aed90d44a4239d5a6e558fe6836da276c/test/StableToken.t.sol#L471) when develop and McMint are merged.

Issue #53, opened 2022-11-01
Implement trading limits in the Broker


bowd:
PoC Implementation: https://github.com/celo-org/celo-monorepo/pull/9895

First version should focus on functionality vs gas usage

philbow61:
#65

PR #65, opened 2022-11-15
feat(broker): trading limits

bowd:
### Description

This PR implements trading limits for the broker via the `TradingLimits` library.
This library provides two structs:
- `TradingLimits.State` which holds the counters associated with a trading limit setup
- `TradingLimits.Config` which holds the configuration for a trading limit setup.

A trading limit is applied to an asset in a pair. For example, say we have the `cUSD/CELO` pair which has an `exchangeId`, if we want to apply a trading limit for cUSD in that pair we configure the trading limits on `limitId = exchangeId ^ cUSDTokenAddress`. Inside of a TradingLimits.Config we can enable 3 trading limits:
- L0 which is a time-window based limit, allowing a netflow of `limit0` tokens every `timespan0` seconds.
- L1 which is the same as L0 but with a different limit and timespan.
- LG which is a global limit without a timespan

### Other changes

- Create a MockExchangeProvider to simplify testing in the Broker
- Compile warning fixes

### Tested

- Library has test
- Integration with broker tested

### Related issues

- #53 

### Backwards compatibility

N/A

### Documentation

N/A


philbow61 (commented):
cool stuff!!!
- enjoyed the usage of the bit mask for the different trading limit variants 
- liked the Balance snapshot and the overall refactoring of the broker test

test/common/TradingLimits.t.sol
```diff
@@ -0,0 +1,274 @@
...
+
+  function test_verify_withLG_andMetNegatively() public {
+    state.netflowGlobal = -1231;
+    vm.expectRevert(bytes("LG Exceeded"));
+    state.verify(configLG(1230));
+  }
+
+  /* ==================== State#update ==================== */
+
+  function test_update_withNoLimit_updatesOnlyGlobal() public {
```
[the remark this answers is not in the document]
bowd: Good catch! At some point I was toying around with the idea of always updating `netflowGlobal` in order to actually be "global", so also when the limit isn't active, but in the end, I decided to keep it simple and only update `netflow*` when the limit is active.

test/common/TradingLimits.t.sol
```diff
@@ -0,0 +1,274 @@
...
+
+  uint8 private constant L0 = 1; // 0b001
+  uint8 private constant L1 = 2; // 0b010
+  uint8 private constant LG = 4; // 0b100
+
+  TradingLimits.State private state;
+
+  function setUp() public {
+    TradingLimits.State memory _state;
+    state = _state;
```
bayological, line 24: Is this setup needed? Surely you get the same result from just declaring it?

bayological (approved):
Nice work on this. LGTM

test/common/TradingLimits.t.sol
bowd: Yeah, I can't reconstruct how I ended up in this situation 😂. Removed it.

PR #68, opened 2022-11-15
feat(broker): make broker capable of minting

### LINKED ISSUE #32: Make the Broker capable of minting StableTokens

philbow61:
### Description

- added the `BROKER_REGISTRY_ID` to the StableToken contract 
- both `mint()` and `burn()` now also check whether the caller is the broker address from the registry
- thereby allowing a registered broker to mint and burn 

### Tested

- updated the McMintIntegration.sol
- The migrated StableToken.t.sol is currently on develop. Therefore I would suggest writing the same test for [mint](https://github.com/mento-protocol/mento-core/blob/dedb150aed90d44a4239d5a6e558fe6836da276c/test/StableToken.t.sol#L183) and [burn](https://github.com/mento-protocol/mento-core/blob/dedb150aed90d44a4239d5a6e558fe6836da276c/test/StableToken.t.sol#L471) when develop and McMint are merged. 

### Related issues

- Fixes #32


test/utils/McMintIntegration.sol
```diff
@@ -53,6 +53,7 @@ contract McMintIntegration is Test, WithRegistry {
   address cUSD_USDCet_oracleReportTarget;
   address cEUR_USDCet_oracleReportTarget;
   address cUSD_cEUR_oracleReportTarget;
+  address exchange;
```
ninocomputer, line 56: why are you using exchange instead of broker?

ninocomputer (changes requested):
looks good, but I wish there were unit tests to verify that changes in the contract work as expected.

test/utils/McMintIntegration.sol
philbow61: Before my branch the `exchangeIdentifier` of the stableToken was set to "Broker" this workaround allowed the broker to mint. Since the Broker is now capable of minting by itself I removed the "Broker" and replaced it with "Exchange"

philbow61:
> looks good, but I wish there were unit tests to verify that changes in the contract work as expected.

@ninocomputer I tried to address the unit test in the description of this PR. The stableToken.t.sol is not yet on this branch that's why I would suggest we add them to the existing mint and burn stable token tests as soon as the branches are merged. But maybe there is a better way of doing this?

ninocomputer:
> > looks good, but I wish there were unit tests to verify that changes in the contract work as expected.
> 
> @ninocomputer I tried to address the unit test in the description of this PR. The stableToken.t.sol is not yet on this branch that's why I would suggest we add them to the existing mint and burn stable token tests as soon as the branches are merged. But maybe there is a better way of doing this?

@philbow61 you can try cherry pick StableToken test commit on your branch

ninocomputer (approved):
looks good! 👍🏼

PR #74, opened 2022-11-22
feat(broker): add burnstabletokens method

### LINKED ISSUE #58: Allow stable burning via Broker
### Description

We're burning cStables via the burning-bot, which takes a long time and has an impact on Mento pools. The collateral we receive from the trading sessions goes back into the reserve. Ideally, we would have a way of directly burning stables instead.

For this, we can add a function `burnStableTokens(address token, uint256 amount)` to the Broker.

### Acceptance criteria 

- Add the permissionless `burnStableTokens(address token, uint256 amount)` function to the Broker
- Write tests for the functionality

ninocomputer:
### Description

In this PR I'm introducing changes in `Broker.sol` making it possible to burn stables without permissions

### Other changes

n/a

### Tested

Unit tests to check everything works as expected

### Related issues

- Fixes #58 

### Backwards compatibility

should be backwards compatible

### DIFF
--- a/test/Broker.t.sol
+++ b/test/Broker.t.sol
@@ -29,6 +29,7 @@ contract BrokerTest is Test {
   event ExchangeProviderAdded(address indexed exchangeProvider);
   event ExchangeProviderRemoved(address indexed exchangeProvider);
   event ReserveSet(address indexed newAddress, address indexed prevAddress);
+  event TokenBurned(address token, uint256 amount);
 
   address deployer = actor("deployer");
   address notDeployer = actor("notDeployer");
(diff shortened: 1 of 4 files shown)


test/Broker.t.sol
```diff
@@ -33,6 +33,7 @@ contract BrokerTest is Test {
   event ExchangeProviderAdded(address indexed exchangeProvider);
   event ExchangeProviderRemoved(address indexed exchangeProvider);
   event ReserveSet(address indexed newAddress, address indexed prevAddress);
+  event TokenBurned(address token, uint256 amount);
```
bowd, line 36: I don't think we need this event. Burns are colloquially marked by a `Transfer` to `address(0)`, which is emitted by the `StableToken` contract so I don't think this event really adds much value. I'd remove it.

bowd (changes requested):
Looks good, but would remove the event, you can check for `Transfer` event in the test.

test/Broker.t.sol
ninocomputer: 👍🏼

bowd (approved):

PR #78, opened 2022-11-28
feat(sdk): create npm package

### LINKED ISSUE #12: Create NPM package for core types
**Description**
The Mento SDK will need access to the interfaces defined in the Mento core project. To keep these in sync easily we need to have these types generated automatically when they are updated in the core repo. Create an NPM mento core npm package that contains typescript types, factories & deployed addresses for each contract.

**Acceptance Criteria**
- [ ] Addresses for each accessible type are available in the package
- [x] Typescript types for each contract is available in the package
- [x] Mento npm organisation

philbow61:
### Description

- created a npm mento organization @mentolabs
- created a npm package that contains typescript types and factories [package](https://www.npmjs.com/package/@mentolabs/core)
- created a workflow that automatically generates the types/factories (generateTypes.js), builds them and publishes them 
- the workflow is triggered by a commit to the feature/mcmint branch in which the contracts/interfaces folder has changed

### Other changes

- in order for the build step to run I had to add @node/node: 18.11.9 to our mento-core devDependencies 
- before an older version of the same package was required by a different dependency 
- this older version caused errors with the typescript package needed for the build step 

### Tested

- in order to test the package I installed the package and interacted with the deployed mcmint contracts on baklava 
- https://github.com/dethcrypto/TypeChain/tree/master/packages/target-ethers-v5

### Related issues

- Fixes #12


tobikuhlmann (changes requested):
Neat PR, Philip! It looks good to me. 

The only thing that I think is missing, already from the issue description, is a license for mento-core and the npm package. It would probably be specified in `package.json` and `package/package.json` like in celo-monorepo: https://github.com/celo-org/celo-monorepo/blob/master/package.json#L5. 

Can you add an Apache 2.0 license if there are no objections to it on Discord? (let's give people time to object until the engineering sync later)

bowd (changes requested):
Hey @philbow61 I know I actually sent you down this path but now that I look at it again I think we made a wrong turn.
I'd like to actually only have one `package.json` at the top of the repo and not use this subfolder approach for building the npm package.

I think we initially did this because we had a failing `postinstall` related to `husky`.
Check the [docs](https://typicode.github.io/husky/#/?id=yarn-2) here on how to deal with that using the `pinst` helper.

Then I'd like the structure to actually be:
```
contracts/ -> the contracts
src/ -> where the typescript types and factories get generated
dist/ -> where the javascript is compiled from the typescript in `src`
```

Also I'd add an `.npmignore` file with at least:
```
.env
test
lib
```

tobikuhlmann (approved):
yup

philbow61:
@tobikuhlmann instead of blacklisting files via the .npmignore we are using the files array within the package.json to explicitly  whitelist files for the package.

MentoProtocolAdmin:
👌

tobikuhlmann:
👌

generateTypes.js
```diff
@@ -0,0 +1,20 @@
...
+  const allFiles = glob(cwd+"/out",files)
+
+  const result = runTypeChain({
+   cwd,
+   filesToProcess: allFiles,
+   allFiles,
+   outDir: 'src',
+   target: 'ethers-v5',
+})
+}
```
bowd, line 19: Code formatting here is a bit weird - first indent is two spaces, 2nd indent is 1 space, the brackets at the end of the block are at the same level. can you clean it up pls!

bowd (changes requested):

PR #82, opened 2022-11-29
feat(reserve): allow limitless transfer of collateral assets

### LINKED ISSUE #79: Update reserve to allow transfer of collateral assets
**Description**
The initial implementation of multi-collateral mento, before migrating to the mento-core repo, contained a function([transferExchangeCollateralAsset](https://github.com/celo-org/celo-monorepo/blob/bdb4f0012e072b3ae302cbdfc91aed054c6a830a/packages/protocol/contracts/stability/Reserve.sol#L569)) that allowed the transfer of reserve collateral assets without taking into account the reserve spending limits. This function was not included in the migration. The current code uses the function [transferCollateralAsset](https://github.com/mento-protocol/mento-core/blob/492801d4b0bd81cec074049066cbf968138fdc7c/contracts/Reserve.sol#L451) which takes into account the daily reserve spending limits. 

**Acceptance Criteria**

-  Broker should use collateral transfer function that does not take spending limits into account.

ninocomputer:
### Description

In this pr, I'm updating a missing functionality in reserve, which will enable the broker to transfer collateral assets without being an object to spending limits

### Tested

unit tests

### Related issues

- Fixes #79 

### Backwards compatibility

n/a

### Documentation

n/a

### DIFF
--- a/contracts/Reserve.sol
+++ b/contracts/Reserve.sol
@@ -492,6 +492,23 @@ contract Reserve is IReserve, ICeloVersionedContract, Ownable, Initializable, Us
   }
 
   /**
+   * @notice Transfer collateral asset to any address.
+   * @dev Transfers are not subject to a daily spending limit.
+   * @param collateralAsset The address of collateral asset being transferred.
+   * @param to The address that will receive the collateral asset.
+   * @param value The amount of collateral asset to transfer.
+   * @return Returns true if the transaction succeeds.
+   */
+  function transferExchangeCollateralAsset(
+    address collateralAsset,
+    address payable to,
+    uint256 value
+  ) external returns (bool) {
+    require(isExchangeSpender[msg.sender], "Address is not allowed to spend");
+    return _transferCollateralAsset(collateralAsset, to, value);
+  }
+
+  /**
    * @notice Transfer unfrozen gold to any address.
    * @param to The address that will receive the gold.
    * @param value The amount of gold to transfer.
--- a/contracts/Reserve.sol
+++ b/contracts/Reserve.sol
@@ -504,7 +504,7 @@ contract Reserve is IReserve, ICeloVersionedContract, Ownable, Initializable, Us
     address payable to,
     uint256 value
   ) external returns (bool) {
-    require(isExchangeSpender[msg.sender], "Address is not allowed to spend");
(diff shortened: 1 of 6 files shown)


contracts/Reserve.sol
```diff
@@ -491,6 +491,23 @@ contract Reserve is IReserve, ICeloVersionedContract, Ownable, Initializable, Us
...
 
+  /**
+   * @notice Transfer collateral asset to any address.
+   * @dev Transfers are not subject to a daily spending limit.
+   * @param collateralAsset The address of collateral asset being transferred.
+   * @param to The address that will receive the collateral asset.
+   * @param value The amount of collateral asset to transfer.
+   * @return Returns true if the transaction succeeds.
+   */
+  function transferExchangeCollateralAsset(
```
bayological, line 502: If there is no need for the function [`transferCollateralAsset`](https://github.com/mento-protocol/mento-core/blob/492801d4b0bd81cec074049066cbf968138fdc7c/contracts/Reserve.sol#L451) then it should be removed and this function should be renamed to **transferCollateralAsset**. This name is a bit simpler and easier to understand.

bayological (changes requested):

contracts/Reserve.sol
bowd: The Reserve still has this logic of having one or more "spenders" which are EOAs or Multisigs that can extract funds from the Reserve, for which we have the `ReserveSpenderMultisig`, and also "exchange spenders" which refers to smart contracts from the protocol which need to spend funds during swaps (previously the Exchange contracts, and now the single Broker contract).

So until we revisit the above completely we will always need two copies of functions that interact with the funds: 
- `transferGold` and `transferCollateralAsset` called by the reserve spenders (with spending limits)
- `transferExchangeGold` and `transferExchangeCollateralAsset` called by the reserve _exchange_ spenders (without spending limits)

Again, there might be some superfluous logic here that we can simplify but for now we should keep it consistent.

bowd (approved):
LGMT! 🚢 

There's just one thing I want you to change, I was trying to figure out why there was no need to change the integration tests to make it work with the new method and realized that I did this in the `IntegrationSetup.sol` file:
```solidity
    reserve.addExchangeSpender(address(broker));
    reserve.addSpender(address(broker));
```

Can you remove the `addSpender` call here, pls.

PR #83, opened 2022-11-30
feat(breaker): make breakers configurable for specific ratefeeds

### LINKED ISSUE #81: Make Breakers configurable for specific rate feeds
### Description

Currently, the BreakerBox has an array of registered breakers. When a new price is reported, the breaker iterates through the list and checks each breaker for the `rateFeedID`. This means that there is no way to configure a breaker that's only checked for one `rateFeed`. At the same time, the existing MedianDeltaBreaker has a global `priceChangeThreshold` which is checked for all rates. 

In order to increase how configurable the system is we need to tackle two things:
1. In the `BreakerBox` we need to have a way of specifying if a breaker is enabled for a given `rateFeedID` by storing a `mapping(address => mapping(address => bool)) breakerEnabled` that we can check with `breakerEnabled[breaker][rateFeedID]`. We will then check this mapping when deciding if a breaker should be verified.

2. In the `MedianDeltaBreaker` we want to have a global `uint256 defaultPriceChangeThreshold` and a configurable `mapping(address => uint256) priceChangeThreshold` where we can store the threshold by `rateFeedID`. Then when verifying the median, we use the default or the specific value for the rate feed if it's configured.

Notes:
- We have to treat the edge case when we're disabling a breaker for a rate feed if that breaker is already tripped. We should reset the trading mode of the rate feed to the default.

Side quests:
- Maybe rename `priceChange` to `rateChange` in the `MedianDeltaBreaker` to keep things consistent.
- **If the deployment branch gets merged while this is in-flight** the upgrade scripts need to be updated to reflect the change.

### Acceptance criteria

- [x] BreakerBox has a configurable mapping between breakers and rate feeds that it takes into account
- [x] MedianDeltaBreaker has configurable thresholds per rate feed
- [x] We have tests that reflect general code paths, including but not limited to when a breaker which was tripped is disabled from a price feed.

ninocomputer:
### Description

Currently, the BreakerBox has an array of registered breakers. When a new price is reported, the breaker iterates through the list and checks each breaker for the rateFeedID. This means that there is no way to configure a breaker that's only checked for one rateFeed. At the same time, the existing MedianDeltaBreaker has a global priceChangeThreshold which is checked for all rates. The goal of this pr is to make this system more configurable.

### Other changes

Renamed priceChange to rateChange in the MedianDeltaBreaker and tests to keep things consistent.

### Tested

Unit tests

### Related issues

- Fixes #81 

### Backwards compatibility

n/a

### DIFF
--- a/contracts/BreakerBox.sol
+++ b/contracts/BreakerBox.sol
@@ -32,7 +32,7 @@ contract BreakerBox is IBreakerBox, Initializable, Ownable {
   // Ordered list of breakers to be checked.
   LinkedList.List private breakers;
   // Maps a breaker with rate feed id and bool to check if it's enabled.
-  mapping(address => mapping(address => bool)) breakerEnabled;
+  mapping(address => mapping(address => bool)) public breakerEnabled;
 
   // Address of the Mento SortedOracles contract
   ISortedOracles public sortedOracles;
@@ -140,7 +140,11 @@ contract BreakerBox is IBreakerBox, Initializable, Ownable {
     emit BreakerRemoved(breaker);
   }
 
-  // todo write natspec to describe the changes
+  /**
+   * @notice Sets a breaker that is enabled for the specified rate feed.
+   * @param breaker The address of the breaker to be enabled.
+   * @param rateFeedID The address of the rate feed.
+   */
   function setBreakerEnabled(address breaker, address rateFeedID) public onlyOwner {
     TradingModeInfo memory info = rateFeedTradingModes[rateFeedID];
     require(info.lastUpdatedTime != 0, "this rate feed has not been registered");
@@ -261,6 +265,15 @@ contract BreakerBox is IBreakerBox, Initializable, Ownable {
     return info.tradingMode;
   }
 
+  /**
+   * @notice Checks if a breaker is enabled for a specific rate feed.
+   * @param breaker The address of the breaker we're checking for.
+   * @param rateFeedID The address of the rateFeedID.
+   */
+  function isBreakerEnabled(address breaker, address rateFeedID) external view returns (bool) {
+    return breakerEnabled[breaker][rateFeedID];
+  }
+
   /* ==================== Check Breakers ==================== */
 
   /**
@@ -309,7 +322,7 @@ contract BreakerBox is IBreakerBox, Initializable, Ownable {
 
     // Check all breakers.
     for (uint256 i = 0; i < _breakers.length; i++) {
-      if (breakerEnabled[_breakers[i]][rateFeedID]) {
+      if (isBreakerEnabled(_breakers[i], rateFeedID)) {
         IBreaker breaker = IBreaker(_breakers[i]);
         bool tripBreaker = breaker.shouldTrigger(rateFeedID);
         if (tripBreaker) {
--- a/contracts/BreakerBox.sol
+++ b/contracts/BreakerBox.sol
@@ -4,7 +4,6 @@ pragma solidity ^0.5.13;
 import { IBreakerBox } from "./interfaces/IBreakerBox.sol";
 import { IBreaker } from "./interfaces/IBreaker.sol";
 import { ISortedOracles } from "./interfaces/ISortedOracles.sol";
-import { Test, console2 as console } from "celo-foundry/Test.sol";
 
 import { Ownable } from "openzeppelin-solidity/contracts/ownership/Ownable.sol";
 import { AddressLinkedList, LinkedList } from "./common/linkedlists/AddressLinkedList.sol";
@@ -145,12 +144,12 @@ contract BreakerBox is IBreakerBox, Initializable, Ownable {
   }
 
   /**
-   * @notice Enables a breaker for the specified rate feed.
-   * @param breaker The address of the breaker to be enabled.
+   * @notice Enables or disables a breaker for the specified rate feed.
+   * @param breaker The address of the breaker.
    * @param rateFeedID The address of the rate feed.
    * @param status Bool indicating the the status.
    */
-  function setBreakerEnabled(
+  function toggleBreaker(
     address breaker,
     address rateFeedID,
     bool status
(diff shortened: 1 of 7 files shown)


contracts/BreakerBox.sol
```diff
@@ -129,15 +133,53 @@ contract BreakerBox is IBreakerBox, Initializable, Ownable {
...
+    address breaker,
+    address rateFeedID,
+    bool status
+  ) public onlyOwner {
+    TradingModeInfo memory info = rateFeedTradingModes[rateFeedID];
+    require(info.lastUpdatedTime != 0, "this rate feed has not been registered");
+    require(isBreaker(breaker), "this breaker has not been registered in the breakers list");
+    breakerEnabled[breaker][rateFeedID] = status;
+    emit BreakerStatusUpdated(breaker, rateFeedID, status);
+  }
```
bowd, line 181: It doesn't make sense to have both functions if `status` is a variable. And you should name it something like `toggleBreaker`, and have a single function that sets the status.

But also keep note that disabling a breaker requires additional logic to check if the breaker is currently sprung for that `rateFeedId`, which means:
- Getting the tradingModeInfo for the ratefeedId
- Checking if the breaker for that trading mode == breaker
- If yes then reset trading mode

contracts/interfaces/IBreaker.sol
```diff
@@ -18,6 +18,18 @@ interface IBreaker {
...
+   * @param defaultRateChangeThreshold The value of the new threshold.
+   */
+  event DefaultRateChangeThresholdUpdated(uint256 defaultRateChangeThreshold);
+
+  /**
+   * @notice Emitted when the rate feed is configured to a rate threshold.
+   * @param rateChangeThreshold The threshold of the rate feed.
+   */
+  event RateChangeThresholdForRateFeedUpdated(address rateFeedID, uint256 rateChangeThreshold);
+
```
bowd, line 32: These should go in the `MedianDeltaBreaker` because they are events particular to that Breaker implementation, not events that would need to be emitted by all breakers.

contracts/MedianDeltaBreaker.sol
```diff
@@ -125,11 +151,26 @@ contract MedianDeltaBreaker is IBreaker, Ownable {
...
+
+    uint256 rateSpecificThreshold = rateChangeThreshold[rateFeedID].unwrap();
+
+    // checks if a given rate feed id has a threshold set
+    if (rateSpecificThreshold != 0) {
+      allowedThreshold = rateSpecificThreshold;
+    }
+   
+    // otherwise just uses a default threshold
+    allowedThreshold = defaultRateChangeThreshold.unwrap();
```
bowd, line 173: Can you read through this code again and tell me what's wrong with it?

contracts/MedianDeltaBreaker.sol
```diff
@@ -61,13 +63,37 @@ contract MedianDeltaBreaker is IBreaker, Ownable {
...
+      rateFeedIDs.length == rateChangeThresholds.length,
+      "rate feeds and rate change thresholds have to be the same length"
+    );
+    for (uint256 i = 0; i < rateFeedIDs.length; i++) {
+      if (rateFeedIDs[i] != address(0) && rateChangeThresholds[i] != 0) {
+        FixidityLib.Fraction memory _rateChangeThreshold  = FixidityLib.wrap(rateChangeThresholds[i]);
+        require(sortedOracles.getOracles(rateFeedIDs[i]).length > 0, "rate feed ID does not exist as it has 0 oracles");
+        require(_rateChangeThreshold.lt(FixidityLib.fixed1()), "rate change threshold must be less than 1");
+        rateChangeThreshold[rateFeedIDs[i]] = _rateChangeThreshold;
+        emit RateChangeThresholdForRateFeedUpdated(rateFeedIDs[i], rateChangeThresholds[i]);
```
bowd, line 94: I would just name this event `RateChangeThresholdUpdated`, and it would be pretty clear from the arguments that it refers to a single rate feed, and it stands in better contrast with `DefaultRateChangeThresholdUpdated`

test/MedianDeltaBreaker.t.sol
```diff
@@ -138,42 +161,97 @@ contract MedianDeltaBreakerTest_constructorAndSetters is MedianDeltaBreakerTest
...
-    assertTrue(breaker.shouldTrigger(address(rateFeedID)));
+    // rateChangeThreshold not configured
+    updateMedianByPercent(1.2 * 10**24, rateFeedID);
+    assertTrue(breaker.shouldTrigger(rateFeedID));
+
+    // rateChangeThreshold 13 %
+    updateMedianByPercent(1.2 * 10**24, rateFeedIDs[1]);
+    assertTrue(breaker.shouldTrigger(rateFeedIDs[1]));
   }
 }
```
bowd, line 257: I'd like to see tests where the rate feed starts with the default threshold and changes its behavior when its rate-specific threshold is configured. For example:

1. Rate feed ID doesn't have a specific threshold set
2. We create a median change that would result in the default threshold being passed
3. We check that the breaker should trigger
4. We set a rate-specific threshold that's larger than the median change we introduced
5. We check that the breaker doesn't trigger anymore

And the reverse:

1. Rate feed ID doesn't have a specific threshold set
2. We create a median change that would result in the default threshold NOT being passed
3. We check that the breaker doesn't trigger
4. We set a rate-specific threshold that's small enough to result in a trigger with the median change we introduced
5. We check that the breaker starts to trigger

You can also think about the reverse but with disabling the rate-specific threshold by setting it to 0 and flipping the behavior.

contracts/MedianDeltaBreaker.sol
ninocomputer: 👍🏼

contracts/interfaces/IBreaker.sol
ninocomputer: 👍🏼

contracts/BreakerBox.sol
ninocomputer: yes! idk why I left them separate 😩

contracts/MedianDeltaBreaker.sol
ninocomputer: It was supposed to be the other way around

bowd (approved):

PR #93, opened 2023-01-03
Make Broker quote functions external view

### LINKED ISSUE #92: Quote functions in the broker should be external view
#### **Description**

The functions used in the Broker contract to get quotes from McMint (getAmountOut and getAmountIn) only have visibility `external`, which means they can only be call from the outside world as a transaction and not as a contract read call.

#### **Expected result**

getAmountOut/getAmountIn in the broker should be callable as a read contract call

#### **Actual result**

getAmountOut/getAmountIn in the broker can only be called as a transaction

nvtaveras:
### Description

I noticed while working in the SDK that the getAmountOut and getAmountIn functions in the broker can only be called as a transaction due to their visibility, even though they don't modify any state. Instead they should be callable as a contract read call as this is obviously the intended way to be used by consumers (getting a quote -> executing a swap).

### Other changes

N/A

### Tested

No additional testing aside from the existing foundry tests. 

### Related issues

- Fixes #92 

### Backwards compatibility

Not sure if this will require a full redeploy or just a broker upgrade?

### DIFF
--- a/contracts/Broker.sol
+++ b/contracts/Broker.sol
@@ -108,7 +108,7 @@ contract Broker is IBroker, IBrokerAdmin, Initializable, Ownable {
     address tokenIn,
     address tokenOut,
     uint256 amountOut
-  ) external returns (uint256 amountIn) {
+  ) external view returns (uint256 amountIn) {
     require(isExchangeProvider[exchangeProvider], "ExchangeProvider does not exist");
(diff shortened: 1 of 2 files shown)


ninocomputer (approved):

nvtaveras:
Talked to @bayological about this and we will hold on merging this until @bowd is back and we get an update on the state of the audit

bayological (approved):

PR #101, opened 2023-01-25
feat: add workflow for slither

### LINKED ISSUE #95: Add Slither static analysis to CI workflow
**Description**
Create a new action for running Slither against the core contracts and add to the CI job.

**Acceptance Criteria**
- [ ] Slither analyzer runs as part of PR checks

bowd:
### Description

This PR adds Slither to our CI workflows in a "passive" way, this means that it currently just runs and doesn't fail on issues.
In order to get it to run I had to get around [the issues](https://github.com/crytic/slither/issues/1319) that blocked the issue last sprint. This issue presented itself when solc < 0.6 compiled renamed imports like `import { Something as SomethingElse } from "..."`. This semantic is only used in our repo in a dependency of our test suite, therefore I could get around this by skipping test files in compilation running forge as follows (and can be seen in the ci config file):

```bash
forge build --skip test --skip script --build-info
```
But in order to get this to work I also had to rename some of our test files to `.t.sol` because that's what's used as a matcher in this command.

I did a `triage` pass over the slither issues and ignored some that we don't want to fix at the moment, but there are still a few that need to be handled before we can fully enable slither and have it actually stop PRs, this will be done in a separate branch.

### Other changes

- The renaming as explained in the description.

### Tested

Made blocking + not blocking runs in CI, also tested Serif output which is quite nice: https://github.com/mento-protocol/mento-core/runs/10896182685

### Related issues

- Fixes #95 

### Backwards compatibility

it just is.

### Documentation

Updated README


philbow61 (commented):
LGTM!

philbow61 (approved):

README.md
```diff
@@ -38,6 +38,25 @@ forge build
 forge test
 ```
 
+#### Slither
```
tobikuhlmann, line 41: LGTM 👨‍💻  Just one question, should we add this as a dependency such that everybody can run it locally? I had to manually install it with pip

bowd: I wouldn't invest in rigging the repo for python dependencies for this, but I did add a note in the README regarding how to install it.

tobikuhlmann: sounds good to me

tobikuhlmann (approved):

PR #116, opened 2023-01-26
fix: initilize local variables

### LINKED ISSUE #104: Slither fixes for uninitialized-local
**Description**

```
❯ slither . --foundry-ignore-compile --detect uninitialized-local
--ignore-compile used, if something goes wrong, consider removing the ignore compile flag

Reserve.transferCollateralAsset(address,address,uint256).spendingLimitForThisAsset (contracts/Reserve.sol#462) is a local variable never initialized
BreakerBox.removeRateFeed(address).rateFeedIndex (contracts/BreakerBox.sol#203) is a local variable never initialized
Reference: https://github.com/crytic/slither/wiki/Detector-Documentation#uninitialized-local-variables
. analyzed (91 contracts with 1 detectors), 2 result(s) found


```

Fix slither issues for the `uninitialized-local` detector.

**Acceptance criteria**

- [ ] Slither returns no issues

bayological:
### Description

Update to initialize local variables.

### Other changes

- Updated reserve to store remaining collateralAssetSpendingLimit 

### Tested

- Unit tests run and pass as before
- Added tests to cover transfers that hit spending limit for collateral assets

### Related issues

- Closes #104 

### Backwards compatibility

N/A

### Documentation

N/A

### DIFF
--- a/contracts/Reserve.sol
+++ b/contracts/Reserve.sol
@@ -459,7 +459,7 @@ contract Reserve is IReserve, ICeloVersionedContract, Ownable, Initializable, Us
       getDailySpendingRatioForCollateralAsset(collateralAsset) > 0,
       "this asset has no spending ratio, therefore can't be transferred"
     );
-    uint256 spendingLimitForThisAsset;
+    uint256 spendingLimitForThisAsset = 0;
     uint256 currentDay = now / 1 days;
     if (currentDay > collateralAssetLastSpendingDay[collateralAsset]) {
       uint256 balance = getReserveAddressesCollateralAssetBalance(collateralAsset);
--- a/contracts/Reserve.sol
+++ b/contracts/Reserve.sol
@@ -55,6 +55,7 @@ contract Reserve is IReserve, ICeloVersionedContract, Ownable, Initializable, Us
   mapping(address => uint256) public collateralAssetLastSpendingDay;
   address[] public collateralAssets;
   mapping(address => bool) public isCollateralAsset;
+  mapping(address => uint256) public collateralAssetSpendingLimit;
 
   event TobinTaxStalenessThresholdSet(uint256 value);
   event DailySpendingRatioSet(uint256 ratio);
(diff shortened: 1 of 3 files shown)


philbow61 (approved):
marvelous 🚢

ninocomputer (approved):
genius 👍🏼

contracts/Reserve.sol
```diff
@@ -459,7 +459,7 @@ contract Reserve is IReserve, ICeloVersionedContract, Ownable, Initializable, Us
       getDailySpendingRatioForCollateralAsset(collateralAsset) > 0,
       "this asset has no spending ratio, therefore can't be transferred"
     );
-    uint256 spendingLimitForThisAsset;
+    uint256 spendingLimitForThisAsset = 0;
```
bowd, line 462: Huh, this is actually a funny one. This actually is the bug I mentioned briefly in our 1:1.
Compare the implementation of this function to the one in `transferGold`, focus on the usage of `spendingLimit` vs `spendingLimitForThisAsset`. 

This issue needs more refactoring, also @ninocomputer take a look here cose I think you implemented this.

bowd (changes requested):

contracts/Reserve.sol
bayological: @bowd The difference I see is that the `spendingLimit` is declared as a public state variable, whereas `spendingLimitForThisAsset` is function scoped, which aligns with what slither is saying. What was the bug? I don't remember 🥲

bowd: Well, in the `transferGold` logic the `spendingLimit` being a public state var means that the value that is calculated during the execution is actually cached between executions and is reset when the condition – `if (currentDay > lastSpendingDay)` – is met, but that value keeps decreasing during the day.
However in the `transferCollateralAsset` that value is local and initialized at 0 and *is not cached* between runs, which means that `pendingLimitForThisAsset == 0` for most invocations of that functions, expect for when the condition is met – `(currentDay > collateralAssetLastSpendingDay[collateralAsset])` – at which point the limit is *temporarily* reset to the value it should have for the next day, but only for that transaction, then it will be `0` until the next day comes.

bayological: Will resolve in this PR too

bowd (approved):
🙏 BOSS 🇧🇦 🐗 🛥️

PR #136, opened 2023-02-01
feat: add a missing same asset check in BiPoolManager

ninocomputer:
### Description

Added a check in `BiPoolManager.sol` to verify that exchange can't be created with the same assets

### Other changes

n/a

### Tested

Unit tests 

### Related issues

- Fixes [#144](https://github.com/mento-protocol/mento-general/issues/144)

### Backwards compatibility

n/a

### Documentation

n/a


tobikuhlmann (approved):
yup change and test makes sense to me and passes

philbow61 (approved):
LGTM would only suggest moving the test closer to the other tests that test different asset combinations L420

PR #139, opened 2023-02-01
Fix function logic description in BiPoolManager.sol

tobikuhlmann:
### Description

fixes wrong function logic description

### Other changes

no

### Tested

forge tests pass

### Related issues

- Fixes https://github.com/mento-protocol/mento-general/issues/154

### Backwards compatibility

fully just comment change

### Documentation

no docs

### DIFF
--- a/contracts/StableTokenRegistry.sol
+++ b/contracts/StableTokenRegistry.sol
@@ -79,7 +79,7 @@ contract StableTokenRegistry is Initializable, Ownable {
   }
 
   /**
-   * @notice Removes unwamted token instances.
+   * @notice Removes unwanted token instances.
    * @param fiatTicker The currency that is no longer supported.
    * @param index The index in fiatTickers of fiatTicker.
    */
--- a/contracts/StableTokenRegistry.sol
+++ b/contracts/StableTokenRegistry.sol
@@ -79,7 +79,7 @@ contract StableTokenRegistry is Initializable, Ownable {
   }
 
   /**
-   * @notice Removes unwanted token instances.
+   * @notice Removes unwamted token instances.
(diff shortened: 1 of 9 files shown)


philbow61 (approved):

contracts/StableTokenRegistry.sol
```diff
@@ -79,7 +79,7 @@ contract StableTokenRegistry is Initializable, Ownable {
   }
 
   /**
-   * @notice Removes unwamted token instances.
+   * @notice Removes unwanted token instances.
```
ninocomputer, line 82: let's revert this change

ninocomputer (changes requested):

contracts/StableTokenRegistry.sol
tobikuhlmann: I want to leave it in, it does no hurt, and at least it shows as an addressed finding in the final audit report instead of unaddressed as well as correct language in the git history

tobikuhlmann: since you closed the other PR we can just do it here

ninocomputer: `StableTokenRegistry.sol` is removed from main branch and has never been deployed on mainnet. Also lets only address things mentioned in the ticket, in order to keep it contained and straightforward(only because of the audit, otherwise we could do it as a drive-by change)

tobikuhlmann: ok. Why then is the registry in the develop branch here?

ninocomputer: Develop branch used to be McMint, so any changes not related to McMint were merged into the main back in the day

ninocomputer (approved):

PR #140, opened 2023-02-01
refactor: change division, multiplication order

philbow61:
### Description

- changed the order of division and multiplication for calculating the lastUpdate
- this increases precision because division gets truncated  

### Other changes

N/A

### Tested

- calculated whether the new formula is equal to the old one
- all tests still pass

### Related issues

- Fixes https://github.com/mento-protocol/mento-general/issues/150

### Backwards compatibility
N/A

### Documentation

N/A

### DIFF
--- a/contracts/StableToken.sol
+++ b/contracts/StableToken.sol
@@ -490,8 +490,9 @@ contract StableToken is
     FixidityLib.Fraction memory currentInflationFactor = FixidityLib.wrap(numerator).divide(
       FixidityLib.wrap(denominator)
     );
-    uint256 lastUpdated = inflationState.factorLastUpdated.add(inflationState.updatePeriod.mul(timesToApplyInflation));
-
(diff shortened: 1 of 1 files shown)


tobikuhlmann (approved):
math changes seem correct to me and tests pass :)

Just one question: Does it still make sense to define the `uint256 timesToApplyInflation` variable in line 473? It's only used once after this change and I was wondering whether it's more gas efficient to not define it but insert the equation directly into line 480, the only place where it's used. On the other hand readability and understandability are better the way it is.

bowd:
@tobikuhlmann local variables aren't expensive at all, what you want to optimize for is storage variables so when you're reading/writing to the contract storage.

bowd:
We don't really care that much about that code tbh cose we'll remove the inflation feature in the future, but given that we're updating the StableToken implementations anyway with MU01 it's cool to add this as a drive-by.

bowd (approved):

PR #148, opened 2023-02-11
chore: add LICENSE and license everything

ninocomputer:
### Description

added a license template and licensed files.

### Other changes

- Updated the comments and param names in BreakerBox.toggleBreaker to be more readable

### Tested

- Ran unit tests to verify functionality works as expected

### Related issues

- Fixes [#160](https://github.com/mento-protocol/mento-general/issues/160)

### Backwards compatibility

n/a

### Documentation

n/a


tobikuhlmann (approved):
yes

philbow61 (approved):
Lgtm but I feel unsure on which files to add the license maybe check with someone more senior @bowd @bayological, to be sure 👍

bayological:
To avoid confusion, I suggest we limit the scope of this to the contracts that we are deploying, specifically the multi-collateral mento and the updated legacy contracts like reserve.sol. It's important to ensure that the identifiers in the source code match what is actually deployed, to avoid confusion or other issues.

bayological (changes requested):
I suggest we limit the scope of this to the contracts that we are deploying, specifically the multi-collateral mento and the updated legacy contracts like reserve.sol. We should make sure that the identifiers in the source code match what is actually deployed, as any differences could cause confusion or other issues.

bowd:
We shouldn't make any more changes to the smart contract code itself because of the audit.

bayological (approved):

PR #154, opened 2023-02-20
fix: make transferCollateralAsset behaviour consistent with transferGold

bowd:
### Description

When implementing our generalized `transferGold` to `transferCollateralAsset` method we skipped a critical check which constrained the logic only to allow transfers to addresses marked as `isOtherReserveAddress`.

This PR introduces this check back and updates the tests accordingly.

### Other changes

None

### Tested

Tests are updated

### Related issues

N/A

### Backwards compatibility

No, but existing version was never released

### Documentation

N/A


nvtaveras (approved):
🚀

PR #158, opened 2023-02-23
feat: add storage layout workflow

### LINKED ISSUE #157: Storage layout checking for proxy implementations
**Description**

We need a CI action that checks our proxied contracts for storage layout changes that are non-additive. 
This would save us from ever merging code that changes storage in a non-backward compatible way.

Look into:
- https://book.getfoundry.sh/reference/forge/forge-inspect
- https://github.com/Rubilmax/foundry-storage-check
- there's also something in celo-monorepo somewhere

**Acceptance**

- [ ] There's a CI step running that complains if the storage layout is modified

philbow61:
### Description

- added a workflow that compares the storage layout of all proxy contract implementations and reverts PRs if they have changed
- The workflow uses  https://github.com/Rubilmax/foundry-storage-check to compare the storage between prior runs that are saved as artifacts.
- In order to perform this action on each of the proxy implementations the get_proxy_names  job creates a matrix containing all the contracts for which we want to check the storage 
- Once this branch is merged to develop it will fail on the first run since the artifacts don't exist yet and need to be built for the first time. 

### Other changes

- N/A

### Tested

- In order to test this I created the test/new-ci-job branch that branches of the feature/add-storage-layout-check branch.
- After this opened a Pr and made changes to the Reserve.sol that change the storage layout.
-  the workflow detected those changes to the storage layout correctly [Action](https://github.com/mento-protocol/mento-core/actions/runs/4253373514)
- once this branch is merged I will close the test branch again.

### Related issues

- Fixes #157 

### Backwards compatibility

- N/A

### Documentation

- N/A

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,7 +48,7 @@ jobs:
       - name: Generate matrix with all contract names from the proxy directory 
         id: set-matrix
         run: |
-          echo "::set-output name=matrix::$(ls contracts/proxies | sed 's/Proxy.sol//')"
+          echo "::set-output name=matrix::$(ls contracts/proxies | sed 's/Proxy.sol//' | jq -R -s -c 'split("\n") | map(select(length > 0))') )"
     outputs:
       matrix: ${{ steps.set-matrix.outputs.matrix }}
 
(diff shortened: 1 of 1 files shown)


tobikuhlmann (approved):
looks gooddd

bowd (approved):

PR #182, opened 2023-04-15
fix: Take token precision into account in BiPoolManager

bowd:
### Description

The BiPoolManager doesn't handle different token precisions correctly. In the past, it had worked because both its internal bucket calculation and the tokens used in the system had 18 decimals. In this update, we introduce USDC which has 6 decimals. This causes a bug in the system because when managing vAMM bucket sizes the assets need to be represented with the say precision so that 1e18 in the cUSD bucket should equal 1e18 in the axlUSDC bucket. But if we don't account for precision, that's actually very off as `1e18 cUSD = $1` and `1e18 axlUSDC = $1,000,000,000,000` (quite some difference).

In order to tackle this we propose that the internal buckets are always represented in 18 decimal precision, similar to how Curve does it. The system will keep track of precision multipliers for all tokens and use them to transform to and from the internal representation. This multiplier will equal `1^(18 - tokenDecimals)`, thus it will be 1 for tokens with 18 decimals and, for example in the case of USDC, it will be `1e12`. 

Tokens with more than 18 decimals aren't supported (the contract reverts in `createExchange`).

I analyzed two options for keeping track of precision multipliers:
1. Add them to the PoolExchange struct:
  a. Cons: breaking change to the struct, duplication for all tokens in the system, upgrade would require reinitializing the exchanges. 
  b. Pros: colocation of relevant data.
2. Create a global mapping token => precisionMultiplier.
  a. Cons: extra global storage, existing tokens must be set during upgrade, otherwise system could be misconfigured. 
  b. Pros: No breaking change on storage structure, pretty easy to upgrade with the added setter.
  
I decided to go for (2) this means that upgrading will require two transactions, changing the implementation, and configuring the precision retroactively. 

### Other changes

- Refactored some of the test helpers around the pricing modules

### Tested

- Introduced new tests specifically targeting the different precision tokens

### Related issues

- Fixes #[issue number here]

### Backwards compatibility

Pseudo-backward compatible. If deployed anew, external interfaces don't change. If upgrading an existing deployment, it does require extra configuration to not put the system in a broken state.

### Documentation

Not applicable


contracts/BiPoolManager.sol
```diff
@@ -47,6 +47,9 @@ contract BiPoolManager is IExchangeProvider, IBiPoolManager, Initializable, Owna
   // Address of the Mento SortedOracles contract
   ISortedOracles public sortedOracles;
 
+  // Toke precision multiplier for tokens with a precision less than 18.
+  mapping(address => uint256) public tokenPrecisionMultipliers;
```
nvtaveras, line 51: nit: this mapping will also include values for tokens with 18 decimals, so maybe remove `with a precision less than 18.` from the comment? Maybe something like `Token precision multiplier for calculating vAMM buckets`?`

contracts/BiPoolManager.sol
```diff
@@ -133,6 +136,7 @@ contract BiPoolManager is IExchangeProvider, IBiPoolManager, Initializable, Owna
   ) external view returns (uint256 amountOut) {
     PoolExchange memory exchange = getPoolExchange(exchangeId);
     (amountOut, ) = _getAmountOut(exchange, tokenIn, tokenOut, amountIn);
+    return amountOut;
```
nvtaveras, line 139: I like this explicit returns statements, how about also adding it to `getAmountIn`?

philbow61 (approved):
LGTM 🚀

ninocomputer (approved):
good work 👍🏼😎

PR #193, opened 2023-05-19
Update the StableToken implementation to a modern ERC20 (Solidity 0.8.18) and restructuring

### LINKED ISSUE #192: SPIKE: ERC20Permit enabled StableTokens and Solidity 0.8
**Description**

There's an external need from Opera for our stable tokens to support [EIP2612](https://eips.ethereum.org/EIPS/eip-2612) which is an ERC20 extension that can be used for gasless transfers by requiring the users to sign a message, which then would allow an infrastructure provider to either give himself allowance and then call `transferFrom`. 

This requires a few primitives to implement and it would be pretty difficult to retrofit it into our solidity 0.5 contracts. At the same time, if we migrate to using modern open zeppelin it's quite easy to do.

This is an exploration of wether we can come up with a development structure that would allow us to use two solidity versions in parallel, in order to gradually migrate our stack to Solidity 0.8, while at the same time being able to interoperate between the two in tests, essentially have the ability to deploy 0.8 contracts in 0.5 tests and vice-versa.

**Acceptance criteria**

Depending on how things go it can either be:
- An final PR that transitions our repo to this new structure
- An experimental base branch and a list of issues that need to be tackled in order to get this fully functional.

bowd:
### Description

This PR restructures the repository with the aim of:
- Better separating legacy code
- Start using solidity 0.8.18 for new features

While it looks big if you look at the diff numbers, most changes are just the result of restructuring the repo and then having to edit import paths. Inside the contracts folder we now have:
- `common`: contracts copied over from `celo-monorepo` which are required. This section should get thinner with time and we should try to replace it with [@celo/contracts](https://www.npmjs.com/package/@celo/contracts) if possible.
- `interfaces`: interfaces for our contracts these can support all solidity versions (`>=0.5.17 <0.9`) so that they can be imported from both runtimes as needed.
- `legacy`: contracts that we're deprecating but still need around for the time being.
- `libraries`: Solidity libraries
- `oracles`: Oracle and circuit-breaker contracts.
- `proxies`: proxies
- `swap`: Buy-and-sell contracts
- `tokens`: Tokens used in the protocol

In order to interoperate Solidity versions we use the `Factory` contract found in `test/utils/Factory.sol`. If a contract is implemented in `0.8.18` it needs an interface that's `>=0.5.17` and then it can be used like this example take from the `BrokerIntegration` tests:
```solidity
  cEURToken = IStableTokenV2(factory.create("StableTokenV2", abi.encode(false)));
  cEURToken.initialize(
      "cEUR",
      "cEUR",
      18,
      REGISTRY_ADDRESS,
      FixidityLib.unwrap(FixidityLib.fixed1()),
      60 * 60 * 24 * 7,
      initialAddresses,
      initialBalances,
      "Exchange"
  );
```
The 2nd argument to the `create` method is an encoded version of the constructor arguments for the contract.

### Related Issues
- Closes #192 

### Other changes

- The changes to the structure made the storage layout verification step in CI a bit harder to do generalized so I opted for manually specifying which contracts need to be verified.
- Added the `Arrays.sol` contract we use in the deployments repo as well. Not super happy with this, I think we should create a repo of `mento/common` that we can import in both to include some of the shared utils we have.

### Tested

- Ran integration tests with the new version of the StableToken contract
- Wrote a fork-test that forks mainnet and upgrades StableToken and tests out the new implementation

### Related issues

N/A

### Backwards compatibility

Yes, the storage layout of SortedOracles is preserved.

### Documentation

We should document that StableToken will support ERC20Permit.


test/legacy/Exchange.t.sol
```diff
@@ -62,7 +62,7 @@ contract ExchangeTest is Test, WithRegistry, TokenHelpers {
     celoToken = new GoldToken(true);
     reserve = new MockReserve();
     exchange = new Exchange(true);
-    stableToken = new StableToken(true);
+    stableToken = IStableTokenV2(factory.create("StableTokenV2", abi.encode(false)));
```
bayological, line 65: Why are we initializing the stableToken this way? Is this a workaround for the Solidity version?

test/utils/Factory.sol
```diff
@@ -0,0 +1,39 @@
...
+import { console } from "forge-std/console.sol";
+import { GetCode } from "./GetCode.sol";
+
+
+interface MiniVM {
+  function etch(address _addr, bytes calldata _code) external;
+  function getCode(string calldata _path) external view returns (bytes memory);
+}
+
+contract Factory {
```
bayological, line 13: A dev comment on why this should be used would be helpful.

test/tokens/StableTokenV2.t.sol
```diff
@@ -0,0 +1,151 @@
...
+      "" // deprecated
+    );
+    token.initializeV2(
+      broker,
+      validators,
+      exchange
+    );
+  }
+
+  function test_initializers_disabled() public {
```
bayological, line 56: Let's stick to the naming convention for these tests, it adds more context and makes it easier to identify the intention of the test and expected result at a glance.
image
vs
image

bayological (changes requested):
All looks fine to me. Left some comments on naming and comments.

bayological (changes requested):
All looks fine to me. Left some comments on naming and comments.

test/legacy/Exchange.t.sol
bayological: Sorry, I've just seen the note on the PR. 👍🏿

bayological (approved):

ninocomputer (approved):

ninocomputer (approved):

nvtaveras (approved):
🚀 🌑

PR #194, opened 2023-05-19
feat: fuzz pricing modules

### LINKED ISSUE #188: Add fuzz tests for pricing modules
**Description**
Verilog provided us with some echidna tests for different components. We were able to integrate all of them except for the pricing module tests. This ticket is about trying to integrate those tests as well. 
The problems i noticed so far: 
- are mainly due to rounding in Wei range
- even with restricting the input of a fuzz run errors still occurred 
`balance = MIN_BALANCE + (balance % (MAX_BALANCE - MIN_BALANCE));` a way to restrict fuzz inputs to a certain range [MIN , Max]

[Tests](https://github.com/Verilog-Solutions/review-mento#automated-testing) 
[echidna notion page](https://www.notion.so/Echidna-Testing-ec0eacab01874abcbf63228c2ea1a62f)

**Acceptance Criteria**

- [ ] Fuzz tests for pricing modules are added and don't fail

bayological:
### Description

This pull request adds fuzz tests for the pricing module contracts as defined in the [Verilog review](https://github.com/Verilog-Solutions/review-mento#automated-testing). The tests verify the consistency and inverse relationship of the getAmountOut and getAmountIn functions in each of the pricing modules.

### Other changes

- N/A

### Tested

These changes have been tested locally using Echidna. Each test checks the inverse relationship between the getAmountOut and getAmountIn functions under a range of different input conditions.

### Related issues

- Closes #188 

### Backwards compatibility

Fully backwards compatible as the changes only involve the addition of tests and do not alter any existing functionality in the smart contracts.

### Documentation

- N/A

### DIFF
--- a/test/echidna/EchidnaConstantProductPricingModule.sol
+++ b/test/echidna/EchidnaConstantProductPricingModule.sol
@@ -1,3 +1,5 @@
+// solhint-disable-next-line max-line-length
+
 pragma solidity ^0.5.13;
 
 import { ConstantProductPricingModule } from "../../contracts/ConstantProductPricingModule.sol";
@@ -6,7 +8,7 @@ import { EchidnaHelpers } from "./EchidnaHelpers.sol";
 
 /**
  * @dev Command for local running:
- * echidna ./test/echidna/EchidnaConstantProductPricingModule.sol --contract EchidnaConstantProductPricingModule --config ./echidna.yaml --test-mode assertion
+ *      echidna ./test/echidna/EchidnaConstantProductPricingModule.sol --contract EchidnaConstantProductPricingModule --config ./echidna.yaml --test-mode assertion
  */
 contract EchidnaConstantProductPricingModule {
   using FixidityLib for FixidityLib.Fraction;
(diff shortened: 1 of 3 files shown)


philbow61 (changes requested):
I think a couple of things need to be changed here:
- right now we are mixing the [boolean-properties](https://github.com/crytic/building-secure-contracts/blob/master/program-analysis/echidna/basic/testing-modes.md#boolean-properties) testing mode with the assertion testing mode. By using function inputs but returning a boolean.
- E.g I changed the return statement to `return ((r+1e18) == amountIn);` the test still pass
- when I added an assert statement `assert(r == amountIn);` the tests fail 
- also I wanted to ask why we are using the between function only for the spread. Based on the pricing module functions I believe the other inputs also need to be restricted since we are using fixidityLib for the calculation.
- Regarding the between function I wanted to ask why we are mapping values outside of the defined range to the max / min value instead of a value in the range. I tried this using this function  ` function _between(uint256 value, unit256 min, uint256 max)public pure returns (uint256) {
        value = min + (value % (max - min));
        return value;
    }`

PR #201, opened 2023-06-05
Feature/refactor breaker box

### LINKED ISSUE #196: Trading mode refactor
**Description**
Modify the trading mode structure in the BreakerBox to enhance its functionality and efficiency. The modification will update the BreakerBox to treat trading modes as binary numbers, where each bit represents a specific trading constraint:
- 2^0 = Outflow trading blocked
- 2^1 = Inflow trading blocked

This approach allows us to aggregate multiple trading modes from different breakers through the use of bitwise OR operations, creating a clear, backwards-compatible method for handling trading modes. 

For this change we should create a new BreakerBox contract, built with a modern solidity version(0.8.18), which will not be upgradeable.

**Tech Notes**

The proposed solution involves refactoring the [`checkAndSetBreakers`](https://github.com/mento-protocol/mento-core/blob/8801ac03c7aa2a03a6e5544eb6ebf71c1fde84c2/contracts/BreakerBox.sol#L315) function to make the code more straightforward and eliminate the potential for future issues due to the current dependency on the ordering of Breakers. 

More info [here](https://www.notion.so/cEUR-cREAL-to-USDC-Swaps-6b9ac311f41e4018a7189c08e789b3b7?pvs=4#59a6a70aadfb48158a90da59488f7ebc)

**Acceptance Criteria**
- [x] TradingModeInfo struct is replaced with BreakerStatus struct
- [x] TradingModeBreaker mapping is removed
- [x] RateFeedTradingMode mapping is removed
- [x] `mapping(address => mapping(address => BreakerStatus)) breakerStatus` added to store the `(breaker, rateFeedID)`
- [x] `mapping(address => uint8) tradingModes` added to store the aggregated `tradingMode` of a `rateFeedID`
- [ ] `checkAndSetBreakers` logic is updated

philbow61:
### Description

- This pr refactors the BreakerBox to support multiple dependent breakers for a rate feed.
- in addition to those changes mentioned in the acceptance criteria I also removed the breakerEnabled mapping since it follows the structure of the new breakerStatus mapping and therefore can be added to the breaker Status struct.
- I use the address(0) breaker to quickly and efficiently track whether a rateFeed is included in the breakerStatus mapping.

### Other changes

- had to update the testAssert as it was dependent on the old mappings in order to get everything to compile
- the only problem with this is that the fork tests fail since the old version is deployed/ used 

### Tested

- updated the BreakerBox tests and added new tests for the changed logic 

### Related issues

- Fixes #196

### Backwards compatibility

- N/A

### Documentation

- N/A


contracts/BreakerBox.sol
```diff
@@ -23,28 +23,20 @@ contract BreakerBox is IBreakerBox, Initializable, Ownable {
...
   address[] public rateFeedIDs;
-  // Maps rate feed ID to its current trading mode info.
-  mapping(address => TradingModeInfo) public rateFeedTradingModes;
-  // Maps a trading mode to the associated breaker.
-  mapping(uint64 => address) public tradingModeBreaker;
+  // Maps a rate feed to it's breakers and their breaker status.
+  mapping(address => mapping(address => BreakerStatus)) public breakerStatus;
+  // Maps a rate feed to the associated trading mode.
+  mapping(address => uint8) public tradingModes;
   // Maps a breaker to the associated trading mode.
```
bayological, line 31: Just for more clarity.
```suggestion 
  // Maps a breaker to the associated trading mode it should activate when triggered.
```

contracts/BreakerBox.sol
```diff
@@ -23,28 +23,20 @@ contract BreakerBox is IBreakerBox, Initializable, Ownable {
...
 
   /* ==================== State Variables ==================== */
-
   address[] public rateFeedIDs;
-  // Maps rate feed ID to its current trading mode info.
-  mapping(address => TradingModeInfo) public rateFeedTradingModes;
-  // Maps a trading mode to the associated breaker.
-  mapping(uint64 => address) public tradingModeBreaker;
+  // Maps a rate feed to it's breakers and their breaker status.
+  mapping(address => mapping(address => BreakerStatus)) public breakerStatus;
```
bayological, line 28: I would consider renaming this to rateFeedBreakerStatus for clarity and consistency.

bayological (changes requested):

contracts/BreakerBox.sol
```diff
@@ -310,55 +278,63 @@ contract BreakerBox is IBreakerBox, Initializable, Ownable {
...
-      }
+  /**
+   * @notice Gets the updated breaker trading mode for a specific rateFeed.
+   * @param rateFeedID The address of the rateFeed.
+   * @param breaker The address of the breaker to update.
+   */
+  function updateBreaker(address rateFeedID, address breaker) internal returns (uint8) {
+    if (rateFeedBreakerStatus[rateFeedID][breaker].tradingMode != 0) {
+      return tryResetBreaker(rateFeedID, breaker);
+    } else return checkBreaker(rateFeedID, breaker);
```
nvtaveras, line 302: no need for the else statement here. also, what happened to your curly brackets in the `else` statements? 😅

contracts/BreakerBox.sol
```diff
@@ -310,55 +278,63 @@ contract BreakerBox is IBreakerBox, Initializable, Ownable {
...
+    BreakerStatus memory _breakerStatus = rateFeedBreakerStatus[rateFeedID][_breaker];
+    IBreaker breaker = IBreaker(_breaker);
+    uint256 cooldown = breaker.getCooldown(rateFeedID);
+    if ((cooldown > 0) && (cooldown.add(_breakerStatus.lastUpdatedTime) <= block.timestamp)) {
+      if (breaker.shouldReset(rateFeedID)) {
+        rateFeedBreakerStatus[rateFeedID][_breaker].tradingMode = 0;
+        rateFeedBreakerStatus[rateFeedID][_breaker].lastUpdatedTime = uint64(block.timestamp);
+        emit ResetSuccessful(rateFeedID, _breaker);
+      } else emit ResetAttemptCriteriaFail(rateFeedID, _breaker);
+    } else emit ResetAttemptNotCool(rateFeedID, _breaker);
```
nvtaveras, line 320: can you add both opening and closing brackets for all the else statements?

contracts/BreakerBox.sol
```diff
@@ -310,55 +278,63 @@ contract BreakerBox is IBreakerBox, Initializable, Ownable {
...
+  /**
+   * @notice Tries to reset a breaker if the cooldown has passed.
+   * @param rateFeedID The address of the rateFeed to run checks for.
+   * @param _breaker The address of the breaker to reset.
+   */
+  function tryResetBreaker(address rateFeedID, address _breaker) internal returns (uint8) {
+    BreakerStatus memory _breakerStatus = rateFeedBreakerStatus[rateFeedID][_breaker];
+    IBreaker breaker = IBreaker(_breaker);
+    uint256 cooldown = breaker.getCooldown(rateFeedID);
+    if ((cooldown > 0) && (cooldown.add(_breakerStatus.lastUpdatedTime) <= block.timestamp)) {
```
nvtaveras, line 314: I think it'd be good to keep this comment from the old code:
`// If the cooldown == 0, then a manual reset is required.`.

Also, I always found `(cooldown.add(_breakerStatus.lastUpdatedTime) <= block.timestamp)` more confusing than `now >= cooldown + lastUpdatedTime`. Not sure if others agree, if yes then maybe we can change it?

contracts/BreakerBox.sol
```diff
@@ -83,121 +68,107 @@ contract BreakerBox is IBreakerBox, Initializable, Ownable {
...
     }
-    delete tradingModeBreaker[tradingMode];
     delete breakerTradingMode[breaker];
-    breakers.remove(breaker);
-
+    uint256 lastIndex = breakers.length.sub(1);
+    if (breakerIndex != lastIndex) {
+      breakers[breakerIndex] = breakers[lastIndex];
+    }
+    breakers.pop();
```
nvtaveras, line 108: this logic looks fine to me, but I didn't see unit tests for calling this function when there's more than one breaker?

nvtaveras (changes requested):
Requesting mostly small changes, it looks great overall!

contracts/BreakerBox.sol
```diff
@@ -103,101 +94,95 @@ contract BreakerBox is IBreakerBox, Initializable, Ownable {
...
-   *      because if its tripped and if we remove it rateFeed will be stuck in the trading mode.
    */
   function removeBreaker(address breaker) external onlyOwner {
     require(isBreaker(breaker), "This breaker has not been added");
 
-    uint64 tradingMode = breakerTradingMode[breaker];
-
-    // Set any refenceRateIDs using this breakers trading mode to the default mode.
-    // Disable a breaker on this address
     address[] memory activeRateFeeds = rateFeedIDs;
```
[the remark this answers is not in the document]
philbow61: I think we can't set the trading mode of all RateFeeds using the breaker to the default trading mode because there could be additional breakers active for that rate feed that are not in the default trading mode.

contracts/BreakerBox.sol
philbow61: i agree 🤝

bayological (approved):

nvtaveras (changes requested):

nvtaveras (approved):
great work! 🚀 🚀 🚀 🚢 🚢 🚢

PR #203, opened 2023-06-07
Fix/expect emit new foundry version

### LINKED ISSUE #202: Refactor tests that use expectEmit cheatcode
**Description**

A recent [update](https://github.com/foundry-rs/foundry/pull/4920) to Foundry has altered the behaviour of the ```expectEmit``` cheatcode, which is breaking our current test suite. In the past, ```expectEmit``` could be called at any point before the function that would cause the event to be emitted. However, the recent update requires that ```expectEmit``` is called immediately before the triggering function.

Our existing tests that are checking for emits now fail due to this change, in order to fix our test suite we need to update all tests to place ```expectEmit``` immediately before the respective call that triggered the event.

**Acceptance Criteria**

- Refactor tests to ensure that ```expectEmit``` is invoked immediately before the call that triggers the event emission
- All tests pass

philbow61:
### Description

The new foundry version changes the logic of how expect emits are evaluated. This caused old tests to fail. 

### Other changes

- N/A

### Tested

- All tests are green again 

### Related issues

- Fixes #202

### DIFF
--- a/test/Exchange.t.sol
+++ b/test/Exchange.t.sol
@@ -443,7 +443,7 @@ contract ExchangeTest_sell is ExchangeTest_stableActivated {
     assertEq(tradableCelo, updatedCeloBucket + amount);
   }
 
-  function test_sellCelo_whenBucketsStaleandReportStale_doesNotUpdateBuckets() public {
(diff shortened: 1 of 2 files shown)


bayological (approved):
Nice work 🔥

Issue #189, opened 2023-05-12
Implement Exponential Moving Average in MedianDeltaBreaker


bayological:
**Description**

This task is to enhance the performance of the MedianDeltaBreaker by implementing an exponential moving average (EMA) mechanism. The intention is to enable the MedianDeltaBreaker to maintain a running average of previous medians which should improve efficiency.

The EMA mechanism will utilize a defined 'smoothing factor' that will determine the weight of the most recent median in the moving average. The weights of previous medians will decay exponentially, making the system more responsive to recent changes.

A new mapping will be introduced to keep track of smoothing factors for each rateFeedID. The first time the function is updated for a particular rateFeedID, it should be set to the current median without using the smoothing factor.

**Tech Notes**

The current logic looks like this:
```
shouldTrigger(rateFeedID):
	currentMedian = sortedOracles.medianRate(rateFeedID)
	previousMedian = previousMedianRates[rateFeedID]
	previousMedianRates[rateFeedID] = currentMedian 

	if previousMedian == 0: return false // first call
	return exceedsThreshold(previousMedian, currentMedian, rateFeedID)
```

With the introduction of the smoothing factor the proposed logic would look like this:
```
shouldTrigger(rateFeedID):
	currentMedian = sortedOracles.medianRate(rateFeedID)
	previousMedian = previousMedianRates[rateFeedID]
	
	if previousMedian == 0: // first call
		previousMedianRates[rateFeedID] = currentMedian
		return false

  smoothingFactor = smoothingFactors[rateFeedID]
	previousMedianRates[rateFeedID] = 
		currentMedian * smoothingFactor + 
		previousMedain * (1-smoothingFactor)
	return exceedsThreshold(previousMedian, currentMedian, rateFeedID)
```

More information can be found [here](https://www.notion.so/cEUR-cREAL-to-USDC-Swaps-6b9ac311f41e4018a7189c08e789b3b7?pvs=4#bb5e9839a7f64623b09df997a2c9dd0c)

**Acceptance Criteria**

- Implement Exponential Moving Average (EMA) in MedianDeltaBreaker.
- Introduce a mapping to hold the smoothing factors for each rateFeedId.
- Add functions to allow smoothing factors to be updated by owner
- Smoothing factor for a rateFeedId should default to 1 if not set.
- The first update for a rateFeedId should set the value to the current median without using the smoothing factor

PR #204, opened 2023-06-08
feat: Use EMA in MedianDeltaBreaker

nvtaveras:
### Description

This implements an exponential moving average calculation for median rates in medianDeltaBreaker. The idea behind this is to be less sensitive to immediate price changes in the median as long as the trend of the price stays within acceptable boundaries. Specs of this feature are documented [here](https://www.notion.so/cEUR-cREAL-to-USDC-Swaps-6b9ac311f41e4018a7189c08e789b3b7?pvs=4#bb5e9839a7f64623b09df997a2c9dd0c).

### Other changes

N/A

### Tested

- Kept the old tests that directly update the value of the EMA which essentially checks that the trigger works.
- Wrote new tests for the EMA calculation and the breaker trigger for both median jumps and drops.

### Related issues

- Fixes https://github.com/mento-protocol/mento-core/issues/189

### Backwards compatibility

- They are compatible since the defaultSmoothingFactor is set to 1 which essentially keeps the same functionality as of today until the smoothingFactor is changed for a rate feed.

### Documentation

N/A


philbow61 (approved):
LGTM 🚢 
enjoyed those well-structured tests 🔥

bayological (approved):

PR #210, opened 2023-06-16
feat: value based constant sum pricing module

### LINKED ISSUE #206: Build a value based constant sum pricing module
**Description**

[TODO] - To be updated once final implementation is clear.

The current constant sum pricing module is doing a token amount based constant sum.
Meaning: 
`Amount(bucket0) + Amount(bucket1) = K ` with K being the constant sum. 
This means that basically `amountIn = amountOut `.
 
This ticket is about changing the logic to calculate the constant sum based on the value in both buckets. 
In order to do that we will need to pass the bucket sizes into the calculation.

**Base Formular:**
`X*p + Y = K ` with: 
`X :=amount of bucket0`
`p :=price of one X in Y` 
`Y :=amount of bucket1`
`K := combined value`

For a trade of `x := amountIn` and `y := amountOut` this means 
`(X+x)*p + Y-y = K`
`=> (X+x)*p + Y - K = y`
`=> (X+x)*p + Y - (X*p +Y) = y`
`=> X*p + x*p - X*p  = y`
`=> x*p = y`

Since the BipoolManager uses the exchange rate to calculate the bucket sizes we can use the ratio of the bucket sizes for our `price := p`.
Additionally, we also want to apply the spread to each swap.

**Acceptance Criteria**

- [x] Value-based Constant Sum Pricing Module is created & tested

philbow61:
### Description

This PR changes the implementation of the constant sum pricing module from token amount constant sum to value-based constant sum. This is achieved by taking the bucket ration as a conversion rate 
`X = TokenInBucketSize`
`Y = TokenOutBucketSize`
`x = amountIn`
`y = amountOut`
`base Formula: (X + x*(1-spread)) * p + Y - y = K`
`with p: Y/X`

`Formula getAmountOut: amountOut = (1 - spread) * amountIn * tokenOutBucketSize) / tokenInBucketSize`
`Formula getAmountIn: amountIn = (amountOut  * tokenInBucketSize) / (tokenOutBucketSize * (1 - spread))`

### Other changes

**BiPoolManager 1:**
- buckets are not updated after each swap for constant sum pairs 
=> price stays constant in between bucket updates 
- quotes/swaps to a constant sum pair fail whenever there is any problem with the current median 

**BiPoolManager 2:**
Problem: 
There was an additional bug in the current implementation:
- the decimal conversion happened before we applied the amounts to the virtual buckets 
- this would have caused problems for Constant Product pairs with different decimals since the pool sizes are 18 decimal based

Solution
=> Amounts are converted to 18 decimals on entry to the BiPoolManger 
=> The BiPoolManager threats everything 18 decimal based internally 
=> Before returning values the BiPoolManger does the decimal conversion back to the token-specific decimals 

### Tested

- new ConstantSum unit tests 
- new BipoolManger unit tests 
- new ConstantSum integration tests

### Related issues

- Fixes #206 

### Documentation
[Notion Page](https://www.notion.so/ConstantSum-Problem-e7a9cc4cc3bd4ff5aa2577f85373daa4)


ninocomputer (changes requested):
just nits

contracts/ConstantSumPricingModule.sol
```diff
@@ -19,51 +19,60 @@ contract ConstantSumPricingModule is IPricingModule {
...
+    uint256 amountOut = numerator.unwrap().div(denominator.unwrap());
+    require(amountOut <= tokenOutBucketSize, "amountOut cant be greater then tokenOutBucketSize");
     return amountOut;
   }
 
   /**
    * @notice Calculates the amount of tokens that should be provided in order to receive the desired amount out.
-   * @dev amountIn = amountOut / (1 - spread)
+   * @dev amountIn = (amountOut  * tokenInBucketSize) / (tokenOutBucketSize * (1 - spread))
+   * @param tokenInBucketSize The bucket size of the token swapt in.
```
ninocomputer, line 54:
```suggestion
   * @param tokenInBucketSize The bucket size of the token swapped in.
```

contracts/ConstantSumPricingModule.sol
```diff
@@ -19,51 +19,60 @@ contract ConstantSumPricingModule is IPricingModule {
...
    * @param tokenOutBucketSize The bucket size of the token swapt out.
```
ninocomputer, line 55:
```suggestion
   * @param tokenOutBucketSize The bucket size of the token swapped out.
```

contracts/ConstantSumPricingModule.sol
```diff
@@ -19,51 +19,60 @@ contract ConstantSumPricingModule is IPricingModule {
...
    * @param spread The spread that is applied to a swap.
    * @param amountOut The amount of tokens in wei that should be swapt out.
```
ninocomputer, line 57:
```suggestion
   * @param amountOut The amount of tokens in wei that should be swapped out.
```

contracts/ConstantSumPricingModule.sol
```diff
@@ -19,51 +19,60 @@ contract ConstantSumPricingModule is IPricingModule {
   /* ==================== View Functions ==================== */
   /**
    * @notice Calculates the amount of tokens that should be received based on the given parameters
-   * @dev amountOut = (1 - spread) * amountIn
+   * @dev amountOut = (1 - spread) * amountIn * tokenOutBucketSize) / tokenInBucketSize
+   * @param tokenInBucketSize The bucket size of the token swapped in.
    * @param tokenOutBucketSize The bucket size of the token swapt out.
```
ninocomputer, line 24:
```suggestion
   * @param tokenOutBucketSize The bucket size of the token swapped out.
```

ninocomputer (changes requested):

sissnad:
Are we testing 1 Wei amount swaps somewhere. I think it would make sense to have it for every pricing module.
1. Price: 2 InAmount: 1 Wei => OutAmount: 0 Wei
2. Price: 0.5 InAmount: 1 Wei => OutAmount: 1 Wei

philbow61:
> Are we testing 1 Wei amount swaps somewhere? I think it would make sense to have this test for every pricing module not only constantSum.
> 
> 1. Price: 2, InAmount: 1 Wei => OutAmount: 0 Wei
> 2. Price: 0.5, InAmount: 1 Wei => OutAmount: 1 Wei

shouldn't it be 
2. Price: 0.5, InAmount: 1 Wei => OutAmount: 1/0.5 = 2 Wei

sissnad:
You right @philbow61. Was a typo

contracts/swap/BiPoolManager.sol
```diff
@@ -475,15 +475,24 @@ contract BiPoolManager is IExchangeProvider, IBiPoolManager, Initializable, Owna
    * @return shouldUpdate
    */
   function shouldUpdateBuckets(PoolExchange memory exchange) internal view returns (bool) {
-    (bool isReportExpired, ) = sortedOracles.isOldestReportExpired(exchange.config.referenceRateFeedID);
+    bool hasUsableMedian = oracleHasUsableMedian(exchange);
```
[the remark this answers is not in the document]
bowd: Let's go for "valid"

test/integration/ConstantSumIntegration.t.sol
```diff
@@ -0,0 +1,191 @@
...
+    IERC20(tokenIn).approve(address(broker), amountIn);
+
+    // Execute swap
+    if (revert) {
+      vm.expectRevert("no usable median");
+    }
+    actualOut = broker.swapIn(address(exchangeProviders[0]), poolId, tokenIn, tokenOut, amountIn, 0);
+  }
+
+  function test_swap_whenConstantSum_pricesShouldStayTheSameInBetweenBucketUpdates() public {
```
bowd, line 63: Should we also test cEUR/USDC? Feels like when the oracle price is not `1` is also an important integration test.

test/swap/BiPoolManager.t.sol
```diff
@@ -886,3 +976,101 @@ contract BiPoolManagerTest_bucketUpdates is BiPoolManagerTest_withExchange {
...
+    changePrank(broker);
+    vm.expectRevert("no usable median");
+    biPoolManager.swapIn(exchangeId_cEUR_bridgedUSDC, address(cEUR), address(bridgedUSDC), 1e18);
+    vm.expectRevert("no usable median");
+    biPoolManager.swapOut(exchangeId_cEUR_bridgedUSDC, address(cEUR), address(bridgedUSDC), 1e18);
+  }
+
+  function test_quotesAndSwaps_whenOldestReportExpired_shouldRevert() public {
+    address USDUSDC_rateFeedID = address(uint160(uint256(keccak256(abi.encodePacked("USDUSDC")))));
+    sortedOracles.setOldestReportExpired(USDUSDC_rateFeedID);
```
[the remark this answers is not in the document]
philbow61: In the unit tests the minimum reports were set explicitly in the pool creation based on what is supposed to be tested. 
Most of the time 0 is used in other cases where the value is essential for the test case it is set to 5.
The same goes for the number of reports. Also, we are using a mock sorted oracles contract here where we can e.g explicitly define the return value of `sortedOracles.isOldestReportExpired()` independent of any rates.

Similar test scenarios are also tested in the ConstantSumIntegrationTests. Here we are using the actual contracts no mocks. Here the pools are created with `minimumReports = 5;` and ten reports are submitted to the sortedOracles contract. Those are then manipulated(e.g) for specific scenarios like `sortedOracles.numRates()` or `sortedOracles.isOldestReportExpired()`

test/integration/ConstantSumIntegration.t.sol
sissnad: Yes, I agree that's important. It is happening as part of the 1 Wei amount test I suggested above. But does not harm to have an individual test case

sissnad: I think this test also test swaps at prices `!= 1` [test_swap_whenConstantSum_pricesShouldChangeWhenBucketRatioChanges()](https://github.com/mento-protocol/mento-core/blob/5aac8af27487b8298a69b6928576d00102296754/test/integration/ConstantSumIntegration.t.sol#L100)

sissnad: What might be interesting as well is going through all tests and check their assumptions. Whereever we assume `price = 1` and/or `cUSDUSDC` 

- we can either replace assumption like `price = 1` and/or `cUSDUSDC` by a `price !=1` and `cEURUSDC`   

- or loop through `[1, 3.7]` and `cEURUSDC`. 

That way we make sure we are not performing the tests for the special case `price = 1 only

sissnad (changes requested):
Either `test_swap_whenCircuitBreakerActivates_shouldDisableTrading()`in `BrokerIntegration.t.sol` should be extended to cover all relevant cases in which trading is disabled. Or we should setup unit tests or lower level integratiton tests that make sure the breaker sets the trading mode correctly in all cases and test then test that broker reacts correctly to the tradinbg mode for all exchanges.

ninocomputer (approved):
overall, it looks good to me! Contract changes make sense and the tests seem comprehensive.

sissnad (approved):
With breaker and constant sum related integration tests being covered by #219 looks good to me!

contracts/swap/BiPoolManager.sol
```diff
@@ -348,12 +355,17 @@ contract BiPoolManager is IExchangeProvider, IBiPoolManager, Initializable, Owna
...
+        exchanges[exchangeId].bucket0 = exchange.bucket0.sub(amountOut);
+        exchanges[exchangeId].bucket1 = exchange.bucket1.add(amountIn);
+      }
     } else {
-      exchanges[exchangeId].bucket0 = exchange.bucket0.sub(amountOut);
-      exchanges[exchangeId].bucket1 = exchange.bucket1.add(amountIn);
+      exchanges[exchangeId].bucket0 = exchange.bucket0;
+      exchanges[exchangeId].bucket1 = exchange.bucket1;
     }
   }
```
bowd, line 370: I would rewrite this like so:

```solidity
if (isConstantProduct) {
  if (tokenIn = exchange.asset0) {
    exchange.bucket0 = exchange.bucket0.add(amountIn);
    exchange.bucket1 = exchange.bucket1.sub(amountOut);
  } else {
    // reverse
  }
}

exchange[exchangeId].bucket0 = exchange.bucket0;
exchange[exchangeId].bucket1 = exchange.bucket1;
if (bucketsUpdated) {
  exchange[exchangeId].lastBucketUpdate = now;
  emit BucketsUpdated(...);
}


```

I think it's a bit clearer and I would add a `@dev` comment in the documentation of the function that explains why we only change the buckets with the amounts when it uses constant product.

contracts/swap/BiPoolManager.sol
```diff
@@ -475,15 +475,29 @@ contract BiPoolManager is IExchangeProvider, IBiPoolManager, Initializable, Owna
    * @return shouldUpdate
    */
   function shouldUpdateBuckets(PoolExchange memory exchange) internal view returns (bool) {
-    (bool isReportExpired, ) = sortedOracles.isOldestReportExpired(exchange.config.referenceRateFeedID);
+    bool hasValidMedian = oracleHasValidMedian(exchange);
+    if (keccak256(abi.encodePacked(exchange.pricingModule.name())) == CONSTANT_SUM) {
```
bowd, line 479: I'm thinking we should add an internal function for this:
```solidity
function pricingModuleIdentifier(PoolExchange memory exchange) internal pure returns (bytes32) {
  return keccak256(abi.encodePacked(exchange.pricingModule.name()));
}

function isConstantSum(PoolExchange memory exchange) internal pure returns (bool) {
  return pricingModuleIdentifier(exchange) == CONSTANT_SUM
}

function isConstantProduct(PoolExchange memory exchange) internal pure returns (bool) {
  return pricingModuleIdentifier(exchange) == CONSTANT_PRODUCT
}
```

From my understanding of how `internal` functions work this shouldn't result in any difference to the bytecode as all internal functions are inlined and it doesn't result in function execution or stacks or anything, but I've never actually looked at the bytecode and I'm quite curious now so I might take a look, either way I'd make this a bit more readable.

test/swap/ConstantSumPricingModule.t.sol
```diff
@@ -16,51 +16,79 @@ contract ConstantSumPricingModuleTest is Test {
...
+    assertEq(constantSum.getAmountIn(10e24, 10e24, 1e23, 0), 0);
+  }
+
+  function test_getAmountIn_whenSpreadIsZero_shouldReturnAmountOutInTokenInValue() public {
+    uint256 tokenInBucketSize = 10e24;
+    uint256 tokenOutBucketSize = 20e24;
+    uint256 spread = 0;
+    uint256 amountOut = 10e18;
+    uint256 amountIn = constantSum.getAmountIn(tokenInBucketSize, tokenOutBucketSize, spread, amountOut);
+    assertEq(amountIn, (amountOut * 1) / 2);
```
bowd, line 65: Is there a reason you're using the multiplication identity operator here? 😂

bowd (changes requested):
Great work! 🎉 🎉 , made some suggestions to improve the readability a bit here and there.

PR #217, opened 2023-06-28
feat: make absolute imports relative

bowd:
### Description

There's some unpleasant behavior when using absolute imports based on remappings and then using this repo as a dependency in `mento-deployments`.
Luckily enough there are only two small places that break this and I changed them in this PR 

### Other changes

N/A

### Tested

Yes

### Related issues

- Refactoring in `mento-deployment`

### Backwards compatibility

Yes

### Documentation

N/A


nvtaveras (approved):

philbow61 (commented):
please ignore slither again

bowd:
Ok so actually it's a new version of slither that's causing these.

PR #223, opened 2023-07-10
Feat/stable token V2 gas functions

### LINKED ISSUE #222: StableTokenV2 gas payment test
**Description**
Investigate/test the StabletokenV2 logic for paying gas in stables compared to the existing implementation. 

**Acceptance Criteria**
- [x] Write tests that verify the logic does the same

philbow61:
### Description

This PR adds the logic for paying gas with stable tokens to the StableTokenV2 contract. 

### Other changes

For testing reasons, I had to add the gas payment functions to the StableToken interfaces. 
Also, I had to change the compiler version for some of the contracts to enable using those together with newer versions. 
Solidity versions are now max 0.8.18 the previous < 0.9 was causing problems with slither and caused the contracts to be compiled in 3 different versions.

### Tested

Wrote unit tests for the V2 gas payments.
StableTokenV1V2GasPayment.t.sol ensures that both implementations do the same. 

### Related issues

- Fixes #222


test/tokens/StableTokenV2.t.sol
```diff
@@ -141,6 +146,50 @@ contract StableTokenV2Test is BaseTest {
...
+  }
+
+  function test_debitGasFees_whenCallerIsVM_shouldDebitGasFees() public {
+    uint256 amount = 100;
+    assertEq(token.balanceOf(holder0), 1000);
+
+    vm.prank(address(0));
+    token.debitGasFees(holder0, 100);
+
+    assertEq(1000 - amount, token.balanceOf(holder0));
```
nvtaveras, line 161: can we put the expected amount on the right for consistency?

test/tokens/StableTokenV2.t.sol
```diff
@@ -141,6 +146,50 @@ contract StableTokenV2Test is BaseTest {
...
+      refund,
+      tipTxFee,
+      gatewayFee,
+      baseTxFee
+    );
+
+    assertEq(1000 + refund, token.balanceOf(holder0));
+    assertEq(tipTxFee, token.balanceOf(feeRecipient));
+    assertEq(gatewayFee, token.balanceOf(gatewayFeeRecipient));
+    assertEq(baseTxFee, token.balanceOf(conmunityFund));
```
nvtaveras, line 190: ditto here, not sure why the expected values here are flipped compare to the tests you wrote in `StableTokenV1V2GasPayment.t.sol`

test/tokens/StableTokenV2.t.sol
```diff
@@ -141,6 +146,50 @@ contract StableTokenV2Test is BaseTest {
...
+  function test_creditGasFees_whenCallerNotVM_shouldRevert() public {
+    vm.expectRevert("Only VM can call");
+    token.creditGasFees(holder0, feeRecipient, gatewayFeeRecipient, conmunityFund, 25, 25, 25, 25);
+  }
+
+  function test_creditGasFees_whenCalledByVm_shouldCreditFees() public {
+    uint256 refund = 25;
+    uint256 tipTxFee = 25;
+    uint256 gatewayFee = 25;
+    uint256 baseTxFee = 25;
```
nvtaveras, line 173: same as in `StableTokenV1V2GasPayment.t.sol`, let's have different values for all of these

nvtaveras (changes requested):
Requested some changes but overall looks good

nvtaveras (approved):
LGTM

PR #226, opened 2023-07-14
test: breaker integration

### LINKED ISSUE #219: Breaker integration test
**Description:**
We need more integration tests that cover more detailed/specific test scenarios.
This issue is about testing breaker-specific logic and scenarios.

**To-do's:**
- create a new BreakerIntegration.t.sol file
- write tests that cover the following scenarios:
- report-> trip -> trade -> fail -> 1 sec warp -> new report -> reset -> trade (All within one block)  
- dependent rate feeds work as expected 
- Talk to Nadiem and Tobi about other tests to write

**Acceptance criteria**
- [x] integration tests for different breaker pool combinations are setup 
- [x] tests are green 🤞

ninocomputer:
### Description

Add integration tests for circuit breaker
Add value delta breaker to integration test setup

### Other changes

n/a

### Tested

Integration tests pass

### Related issues

- Fixes #219 

### Backwards compatibility

n/a

### Documentation

n/a


test/integration/CircuitBreakerIntegration.t.sol
```diff
@@ -0,0 +1,221 @@
...
+
+    // Try swap with shouldBreak true
+    doSwapIn(pair_cUSD_CELO_ID, address(cUSDToken), address(celoToken), true);
+
+    // Check trading mode
+    uint8 tradingMode = breakerBox.getRateFeedTradingMode(cUSD_CELO_referenceRateFeedID);
+    assertEq(uint256(tradingMode), 3); // 3 = trading halted
+
+    // Cool down breaker
+    vm.warp(now + 5 minutes);
```
bowd, line 139: This doesn't actually test what's expected, we need to also do a `setMedianRate` again with a value that still exceeds in order to actually execute the logic that could result in a cooldown, otherwise here you're just validating that the trading mode hasn't changed after 5 minutes without executing any action that could change it.

test/integration/CircuitBreakerIntegration.t.sol
```diff
@@ -0,0 +1,221 @@
...
+    (uint8 medianDeltaTradingMode, , ) = breakerBox.rateFeedBreakerStatus(
+      cUSD_bridgedUSDC_referenceRateFeedID,
+      address(medianDeltaBreaker)
+    );
+    assertEq(uint256(rateFeedTradingMode), 3); // 3 = trading halted
+    assertEq(uint256(valueDeltaTradingMode), 3); // 3 = trading halted
+    assertEq(uint256(medianDeltaTradingMode), 0); // 0 = bidirectional trading
+
+    // Cool down breaker
+    vm.warp(now + 1 seconds);
```
bowd, line 171: Ditto

test/integration/CircuitBreakerIntegration.t.sol
ninocomputer: 👍🏼

test/integration/CircuitBreakerIntegration.t.sol
```diff
@@ -0,0 +1,221 @@
...
+    assertEq(uint256(tradingMode), 0); // 0 = bidirectional trading
+
+    // Try swap with shouldBreak false -> trading is bidirectional again
+    doSwapIn(pair_cUSD_CELO_ID, address(cUSDToken), address(celoToken), false);
+  }
+
+  function test_valueDeltaBreaker_whenMedianExceedsThresholdAndRecovers_shouldHaltAndRecover() public {
+    uint256 blockNumber = block.number;
+    // Trip value delta breaker for cUSD_bridgedUSDC_referenceRateFeedID threshold: 0.1 * 1e24
+    setMedianRate(cUSD_bridgedUSDC_referenceRateFeedID, 1e24 + 1e24 * 0.11);
```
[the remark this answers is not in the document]
ninocomputer: added 2 more tests for it

bowd (approved):

test/integration/CircuitBreakerIntegration.t.sol
```diff
@@ -0,0 +1,296 @@
...
+
+    // Try swap with shouldBreak false -> trading is bidirectional again
+    doSwapIn(pair_cUSD_CELO_ID, address(cUSDToken), address(celoToken), false);
+  }
+
+  function test_medianDeltaBreaker__whenMedianSubceedsThresholdAndRecovers_shouldHaltAndRecover() public {
+    uint256 newMedian = 5e23 - 5e23 * 0.151;
+    // Trip median delta breaker for cUSD_CELO_referenceRateFeedID threshold: 15%
+    setMedianRate(cUSD_CELO_referenceRateFeedID, newMedian);
+    // console.log(5e23 - 5e23 * 0.151, "new threshold");
```
bayological, line 86: Let's get this removed.

test/integration/CircuitBreakerIntegration.t.sol
bayological, line 1: These are good. We should add a test to verify the manual reset logic also, so with both breakers simulate when the breaker has been triggered for rate feeds with 0 cooldown.

test/integration/CircuitBreakerIntegration.t.sol
```diff
@@ -0,0 +1,296 @@
...
+      cUSD_bridgedUSDC_referenceRateFeedID,
+      address(medianDeltaBreaker)
+    );
+    assertEq(uint256(rateFeedTradingMode), 0); // 0 = bidirectional trading
+    assertEq(uint256(valueDeltaTradingMode), 0); // 0 = bidirectional trading
+    assertEq(uint256(medianDeltaTradingMode), 0); // 0 = bidirectional trading
+
+    // Try swap with shouldBreak false -> trading is bidirectional again
+    doSwapIn(pair_cUSD_bridgedUSDC_ID, address(cUSDToken), address(usdcToken), false);
+    assertEq(blockNumber, block.number); // block number should not change
```
bayological, line 152: Why is this assertion needed? In the tests there is no manipulation of the state that would advance the block number.

test/integration/CircuitBreakerIntegration.t.sol
```diff
@@ -0,0 +1,296 @@
...
```
bayological, line 197: Same as above, why is this needed?

bayological (changes requested):

test/integration/CircuitBreakerIntegration.t.sol
philbow61: Yeah, we also didn't like this while writing it. Essentially one of the requirements was to test that the behavior of breaking and recovering works correctly when all happens in the same block.

bayological (approved):

PR #227, opened 2023-07-17
fix: call checkAndSetBreakers in removeReport

philbow61:
### Description

This Pr adds a call to checkAndSetBreakers to the remove report function. Otherwise, there could be a scenario in which removeReport causes a new Median that hasn't been checked by any of the configured breakers and might even trigger a breaker if checked.

### Other changes

- NA

### Tested

- new unit Test that ensures the call to the BreakerBox if Median updated


ninocomputer (approved):
👍🏼

contracts/oracles/SortedOracles.sol
```diff
@@ -399,6 +399,7 @@ contract SortedOracles is ISortedOracles, ICeloVersionedContract, Ownable, Initi
     emit OracleReportRemoved(token, oracle);
     uint256 newMedian = rates[token].getMedianValue();
     if (newMedian != originalMedian) {
+      breakerBox.checkAndSetBreakers(token);
```
bowd, line 402: Let's also add the check that the brakerBox is set before calling this, like in the other function.
We don't have an invariant for this currently so the contract should be operational without a BreakerBox configured.

bowd (changes requested):

bowd (approved):

PR #241, opened 2023-08-02
fix: validate token precision multiplier

### LINKED ISSUE #238: [M-1] setTokenPrecisionMultipliers allows arbitrary decimal multipliers
When the `BiPoolManager` creates an exchange, after performing several validations on the inputs, the token precision multipliers are set in order to deal with assets with different token decimals while calculating swap amounts. 

For example, `USDC` has `6` decimal places, having a token precision multiplier of `1000000000000` (`10 ** (18 - 6)`), as we can see in the function’s logic:

```solidity
tokenPrecisionMultipliers[exchange.asset0] = 10**(18 - uint256(IERC20Metadata(exchange.asset0).decimals()));
tokenPrecisionMultipliers[exchange.asset1] = 10**(18 - uint256(IERC20Metadata(exchange.asset1).decimals()));
```

However, the function `setTokenPrecisionMultipliers` allows to arbitrarily change the token precision multipliers, in this case, say for `USDC`, from `1000000` to an arbitrarily large or small value. Which would then affect the calculations in the swap in/out and in the get amounts in/out functions.

Additionally, the function doesn’t check whether the tokens are valid and registered with the `Reserve` contract. 

**Remediations to Consider**

- Perform the same proper input validations implemented in the `createExchange` function or
- Remove the `setTokenPrecisionMultipliers` from the contract. 🙅‍♂️

**Acceptance Criteria**

- [x] `setTokenPrecisionMultipliers` verifies tokens are valid reserve stable or collateral assets
- [x] `setTokenPrecisionMultipliers` verifies precision multiplier values are in a valid range

bayological:
### Description

This change adds validation to the function `BiPoolManager.setTokenPrecisionMultipliers` similar to that in the create pool function. The change verifies the tokens being added to the mapping are known by the reserve and the precisions are within the allowed range.

### Other changes

- Added an event to be emitted when the multipliers are updated

### Tested

- Unit tests were added to verify the change functions as expected.

### Related issues

- Closes #238 

### Backwards compatibility

- N/A

### Documentation

- N/A


philbow61 (changes requested):
LGTM if we want to go down this route 👍 
I personally would argue there is no need for the setTokenPrecisionMultipliers function in the beginning. 
They are automatically set in the createExchange function by calling decimals() on the token contract. The only context where these multipliers are used is in quotes/swaps and those only work on a valid exchange. So what is the use case for this function in my opinion it can only cause problems by setting wrong values, but maybe I am missing something 😁

bayological:
> LGTM if we want to go down this route 👍 I personally would argue there is no need for the setTokenPrecisionMultipliers function in the beginning. They are automatically set in the createExchange function by calling decimals() on the token contract. The only context where these multipliers are used is in quotes/swaps and those only work on a valid exchanges. So what is the use case for this function in my opinion it can only cause problems by setting wrong values, but maybe I am missing something grin

I agree with you. I don't feel strongly about keeping this function, it does not add much. However, there is no harm in having it there 🤷‍♂️. The only use case that comes to mind is if the multiplier was set incorrectly when creating an exchange, this would allow that to be corrected without recreating the exchange.

philbow61 (approved):
🔥

ninocomputer (approved):

PR #242, opened 2023-08-03
feat: verify pricing modules

### LINKED ISSUE #239: [M-2] Pricing module is not verified when creating exchanges
When creating an exchange in the `BiPoolManager` contract, the pricing module is specified as an input parameter in the `PoolExchange _exchange` struct. However, there are no validations that the address of the pricing module corresponds to a canonical version of this module. This can cause, for example, the previous version of the constant sum module to be used in the exchange creation and, therefore, to calculate swap amounts incorrectly. The only check performed during swaps is in the `pricingModuleIdentifier`:

```solidity
function pricingModuleIdentifier(PoolExchange memory exchange) internal view returns (bytes32) {
  return keccak256(abi.encodePacked(exchange.pricingModule.name()));
} 
```

This also opens up other scenarios where invalid pricing module addresses can be used.

**Remediations to Consider**

Keeping in the `BiPoolManager` contract’s storage the address of the latest or allowed pricing modules and asserting it’s used in `createExchange`.

**Acceptance Criteria**

- [x] BiPoolManager includes a mapping of pricing module identifiers to addresses
- [x] Pricing module is validated on exchange creation

bayological:
### Description

This change adds validation of the pricing module used for exchange creation in `BiPoolManager.createExchange`.

### Other changes
- Added mapping of pricing module identifier to pricing module address.
- Added protected setter function to update the mapping with allowed pricing modules.
- Added call to set pricing modules in integration tests.

### Tested

- Added unit tests to verify functionality works as expected.

### Related issues

- Closes #239 

### Backwards compatibility

- N/A

### Documentation

- N/A


philbow61 (approved):
LGTM 🛥️

contracts/swap/BiPoolManager.sol
```diff
@@ -216,6 +219,21 @@ contract BiPoolManager is IExchangeProvider, IBiPoolManager, Initializable, Owna
...
   }
 
+  /**
+   * @notice Updates the pricing modules for a list of identifiers
+   * @dev This function can only be called by the owner of the contract.
+   *      The number of identifiers and modules provided must be the same.
+   * @param identifiers An array of identifiers for which the pricing modules are to be set.
+   * @param modules An array of module addresses corresponding to each identifier.
+   */
+  function setPricingModules(bytes32[] calldata identifiers, address[] calldata modules) external onlyOwner {
```
ninocomputer, line 229: I don't think just `identifiers` is clear enough maybe I'd change it to `pricingModuleIDs`

ninocomputer (changes requested):

ninocomputer (approved):

contracts/swap/BiPoolManager.sol
philbow61: we decided on not to rename this

PR #254, opened 2023-08-03
feat: add protected median reset function

### LINKED ISSUE #245: [L-3] MedianDeltaBreaker previous rates can not be reset if needed
**Description**
In `MedianDeltaBreaker`, the initial value and start point, where `medianRatesEMA == 0`, only exist the first time a specific `rateFeedID` is set for this breaker. When a `rateFeedID` needs to be removed and later added again into the `MedianDeltaBreaker`, there is no way to reset the initial reference value, which means it cannot easily adjust to newly established rates. It will require a custom higher configuration of the breaker threshold to temporarily bypass the reported values.

**Remediations to Consider**

Implementing a reset function if the breaker is disabled from the `BreakerBox` contract to allow a new initialization of the median rates.

bayological:
### Description

- This change adds a protected function to the median delta breaker that allows the reset of a stored median rate

### Other changes

- None

### Tested

- Added unit tests to verify the functionality

### Related issues

- Closes #245 

### Backwards compatibility

- N/A

### Documentation

- N/A


philbow61 (approved):
LGTM 🛩️

nvtaveras (approved):
🚢

PR #255, opened 2023-08-03
feat: add breaker check in toggleBreaker

### LINKED ISSUE #244: [L-2] toggleBreaker does not update the trading mode for newly added breakers
**Description**
In `BreakerBox.sol`, when turning on the breakers for certain `rateFeedId`s, there will be an execution gap between the `toggleBreaker()` call and the comparison of the corresponding breaker trigger conditions in `checkAndSetBreakers()`. 

Since the `BiPoolManager` contract uses the view function `getRateFeedTradingMode()` before executing swaps, which reads the current breaker status but doesn’t update each dependency:

It is possible that the conditions that trigger a breaker may already be present. However, until the next oracle feed `SortedOracles::report()`, or external call to `checkAndSetBreakers()`, the trading mode won’t consider this breaker. This is especially visible with the `MedianDeltaBreaker`, where the first update will only store the median rate.

**Remediations to Consider**

- Calling `checkAndSetBreakers` when adding breakers to a specific `rateFeedID` in the `toggleBreaker` logic or,
- Make sure to execute a `checkAndSetBreakers` atomically call after enabling a breaker.

philbow61:
### Description

- adds a call to checkAndSetBreakers after enabling breakers for a rate Feed 

### Other changes

- n/a

### Tested

- new unit test that ensures the call is made and trading modes are set after enabling 

### Related issues

- Fixes #244 

### Backwards compatibility

- n/a

### Documentation

- n/a


contracts/oracles/BreakerBox.sol
```diff
@@ -141,6 +141,7 @@ contract BreakerBox is IBreakerBox, Ownable {
     require(rateFeedBreakerStatus[rateFeedID][breakerAddress].enabled != enable, "Breaker is already in this state");
     if (enable) {
       rateFeedBreakerStatus[rateFeedID][breakerAddress].enabled = enable;
+      this.checkAndSetBreakers(rateFeedID);
```
ninocomputer, line 144: I think it'd be better if we changed the `checkAndSetBreakers()` visibility to public since we're calling it internally as well and got rid of the `this.foo()` pattern which required a `call` to be executed and is more expensive.

ninocomputer (changes requested):

ninocomputer (approved):

bowd (approved):

PR #257, opened 2023-08-04
feat: additional validation on tradingLimits

### LINKED ISSUE #240: [L-1] Potential misconfiguration in trading limits
The trading limits act as a safeguard and are checked during the swap operation if the trading config was created for the token. A valid configuration should not permit setting these limits to `0` for enabled directional trading flags. 

The function `Broker.configureTradingLimit()` doesn’t validate the above criteria when setting the trading config for a token. 

In addition, there are no checks to ensure that the trading limit for `L0` is smaller than the trading limit for `L1`, and they’re both smaller than the global trading limit `LG`. 

**Remediations to Consider**

Consider validating the trading config inputs to ensure that invalid trading limits cannot be set.

philbow61:
### Description

- checks for Limit0 < Limit1 < LimitG
- check for Limit0 != 0 and Limit1 != 0  and LimitG != 0 if set 

### Other changes

- n/a

### Tested

- new unit test for validate() function

### Related issues

- Fixes #240 

### Backwards compatibility

- n/a

### Documentation

- n/a


nvtaveras (approved):
Nice! 🚀

PR #259, opened 2023-08-07
feat(MedianDeltaBreaker): add missing access controls

ninocomputer:
### Description

For now, we've decided to ignore the first suggestion and only add the access control checks to the `MedianDeltaBreaker.sol` `shouldTrigger()` to only be callable by BreakerBox

### Other changes

n/a

### Tested

Unit Tests

### Related issues

- 

### Backwards compatibility

n/a

### Documentation

n/a


contracts/oracles/breakers/MedianDeltaBreaker.sol
```diff
@@ -143,6 +159,8 @@ contract MedianDeltaBreaker is IBreaker, WithCooldown, WithThreshold, Ownable {
    *                          should be tripped for the rate feed.
    */
   function shouldTrigger(address rateFeedID) public returns (bool triggerBreaker) {
+    require(msg.sender == breakerBox, "Caller must be the BreakerBox contract");
```
philbow61, line 162: can we move this into a modifier onlyBreakerBox

philbow61 (changes requested):
I think you need to change more that this. The shouldTrigger() function is called as part of the checkAndSetBreaker call in the BreakerBox. Since this function is still public you could still initiate the call to shouldTrigger via a call to checkAndSetBreakers

contracts/oracles/breakers/MedianDeltaBreaker.sol
ninocomputer: I wouldn't add a modifier because we're using this check only in one place.

contracts/oracles/BreakerBox.sol
```diff
@@ -311,6 +311,11 @@ contract BreakerBox is IBreakerBox, Ownable {
    * @param rateFeedID The address of the rateFeed to run checks for.
    */
   function checkAndSetBreakers(address rateFeedID) external {
+    require(
+      msg.sender == address(sortedOracles) || msg.sender == address(this),
+      "Caller must be the SortedOracles or BreakerBox contract"
+    );
+
```
bowd, line 318: I think on the current state of the PR there's no internal call of this function, but it will be when @philbow61 's other PR is merged. I would say to wait for that to be merged and then refactor this a bit into an external `checkAndSetBreakers` which verifies msg.sender, and an internal `_checkAndSetBreakers` which doesn't verify msg.sender, and is the one also called internally.

contracts/swap/BiPoolManager.sol
```diff
@@ -247,14 +247,15 @@ contract BiPoolManager is IExchangeProvider, IBiPoolManager, Initializable, Owna
...
+      revert("asset0 decimals must be <= 18");
+    }
+    if (IERC20Metadata(exchange.asset1).decimals() > 18) {
+      revert("asset1 decimals must be <= 18");
+    }
 
-    tokenPrecisionMultipliers[exchange.asset0] = 10**(18 - uint256(asset0Decimals));
-    tokenPrecisionMultipliers[exchange.asset1] = 10**(18 - uint256(asset1Decimals));
+    tokenPrecisionMultipliers[exchange.asset0] = 10**(18 - uint256(IERC20Metadata(exchange.asset0).decimals()));
+    tokenPrecisionMultipliers[exchange.asset1] = 10**(18 - uint256(IERC20Metadata(exchange.asset1).decimals()));
```
bowd, line 258: Why did you change this here? It's more inefficient to call the `decimals()` view function two times here.

bowd (changes requested):

bowd (approved):

PR #260, opened 2023-08-07
feat(BreakerBox): remove depedencies when removing a rateFeed

ninocomputer:
### Description

- added a fix to remove rateFeedID entry from the rateFeedDependencies mapping when removing the said rate feed from BreakerBox.

### Other changes

n/a

### Tested

Unit test. When trying to access an index that is out of bounds in `rateFeedDependencies(address,uint)` getter it reverts. I didn't want to add another function just for error handling so I added` vm.expecRevert() `clause to test that the entry was removed.

### Related issues

-

### Backwards compatibility

N/A

### Documentation

N/A


philbow61 (approved):
LGTM 🛥️

bowd (approved):

PR #264, opened 2023-08-23
fix: outdated fork-tests after MU03

philbow61:
### Description

This fixes the fork tests which were broken after the many changes done in MU03, mainly the stuff around EMA in the MedianDeltaBreaker, the constantSum pricing module, and the breakerBox refactor.

While working on this we also realized that we forgot to update the breakerBox implementation in the biPoolManager, so god bless the fork tests 🙏 

### Other changes

- Added tests for dependent rate feeds
- Small refactors/added additional comments for context here and here

### Tested

Fork tests are green

> Test result: ok. 18 passed; 0 failed; finished in 116.26s
✨  Done in 129.74s.

### Related issues

Fixes https://github.com/mento-protocol/mento-general/issues/295

### Backwards compatibility

-

### Documentation

-

### DIFF
--- a/test/fork-tests/TestAsserts.t.sol
+++ b/test/fork-tests/TestAsserts.t.sol
@@ -14,6 +14,7 @@ import { IERC20Metadata } from "contracts/common/interfaces/IERC20Metadata.sol";
 import { FixidityLib } from "contracts/common/FixidityLib.sol";
 import { IBreaker } from "contracts/interfaces/IBreaker.sol";
 
+import { BiPoolManager } from "contracts/swap/BiPoolManager.sol";
 import { TradingLimits } from "contracts/libraries/TradingLimits.sol";
 import { WithCooldown } from "contracts/oracles/breakers/WithCooldown.sol";
 import { MedianDeltaBreaker } from "contracts/oracles/breakers/MedianDeltaBreaker.sol";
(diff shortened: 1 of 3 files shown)


test/fork-tests/TestAsserts.t.sol
```diff
@@ -570,23 +591,25 @@ contract TestAsserts is Test {
...
       }
     }
-    if (tradingMode == 1) {
-      (uint256 currentRate, ) = ctx.sortedOracles.medianRate(rateFeedID);
-      return currentRate.add(currentRate.div(1000)); // +0.1%
-    } else if (tradingMode == 2) {
-      return ctx.getValueDeltaBreakerReferenceValue(breaker);
+    bool isMedianDeltaBreaker = breakerIndex == 0;
+    bool isValueDeltaBreaker = breakerIndex == 1;
+    if (isMedianDeltaBreaker) {
```
bowd, line 606: This feels a bit dangerous but we don't have an alternative right now.

test/fork-tests/Utils.t.sol
```diff
@@ -77,6 +77,19 @@ library Utils {
     );
   }
 
+  function getContextForRateFeedID(address _t, address rateFeedID) public view returns (Context memory) {
+    BaseForkTest t = BaseForkTest(_t);
+    (address biPoolManagerAddr, ) = t.exchanges(0);
+    uint256 nOfExchanges = BiPoolManager(biPoolManagerAddr).getExchanges().length;
+    for (uint256 i = 0; i < nOfExchanges; i++) {
+      Context memory ctx = newContext(_t, i);
+      if (getReferenceRateFeedID(ctx) == rateFeedID) {
```
bowd, line 86: We're assuming a 1:1 mapping between rate feed and exchanges here, ok for now but potentially problematic in the future.

bowd (approved):
👍 Looks good. There are some assumptions that feel brittle but I don't see a way around that right now, so I just noted them here. Reading through this I still feel strongly that we should move the Fork Tests to the deployment repo. It would also enable us to more easily do a Simulation + Fork tests run.

ninocomputer (approved):

PR #267, opened 2023-09-08
Feat: Mint function for emission contract

### LINKED ISSUE #305: Governance fork tests

baroooo:
### Description

Instead of pre-minting 100% of the supply, token contract enables emission contract to mint 65% of the total supply.

### Other changes

None

### Tested

Changes are tested with unit tests

### Related issues

- Fixes [#305](https://github.com/mento-protocol/mento-general/issues/305)

### Backwards compatibility

N/A

### Documentation

N/A


contracts/governance/MentoToken.sol
```diff
@@ -10,19 +10,28 @@ import { ERC20Burnable } from "openzeppelin-contracts-next/contracts/token/ERC20
...
    * @param treasuryContract The address of the treasury contract where 10% of the total supply will be sent.
-   * @param emissionContract The address of the emission contract where 65% of the total supply will be sent.
+   * @param emissionContract_ The address of the emission contract where the rest of the supply will be emitted.
    */
   constructor(
     address vestingContract,
     address airgrabContract,
     address treasuryContract,
-    address emissionContract
+    address emissionContract_
```
philbow61, line 34: out of interest why the underscore behind the variable name instead of before 🤔 ? 
In other constructors we used the underscore before so maybe for consistency I would stick to that.

contracts/governance/MentoToken.sol
```diff
@@ -33,12 +42,32 @@ contract MentoToken is ERC20, ERC20Burnable {
...
+  /**
+   * @dev Allows the emission contract to mint new tokens up to the emission supply limit.
+   * @notice This function can only be called by the emission contract and
+   * only if the total emitted amount hasn't exceeded the emission supply.
+   * @param target Address to which the newly minted tokens will be sent.
+   * @param amount Amount of tokens to be minted.
+   */
+  function mint(address target, uint256 amount) external {
+    // Only the emission contract can call this function
+    require(msg.sender == emissionContract, "MentoToken: OnlyEmissionContract");
```
philbow61, line 66: Maybe we can make the revert messages a bit more human-readable and not use CamelCase. I don't have a strong opinion on that, but I would probably do it to be more consistent with other mento contracts.

test/governance/MentoToken.t.sol
```diff
@@ -96,4 +108,60 @@ contract MentoTokenTest is Test {
...
+    vm.expectRevert("MentoToken: OnlyEmissionContract");
+    mentoToken.mint(ALICE, mintAmount);
+  }
+
+  /**
+   * @dev Tests the mint function's behavior when minting amounts that exceed the emission supply.
+   * @notice This test ensures that when the mint function is called with an amount that
+   * exceeds the total emission supply, the transaction should be reverted.
+   */
+  function test_mint_shouldRevert_forAmountsBiggerThanSupply() public {
```
philbow61, line 129:
```suggestion
  function test_mint_shouldRevert_forAmountsBiggerThanEmissionSupply() public {
```

philbow61 (changes requested):

philbow61 (commented):
I think in the past we tried following a different test naming convention:
test_"FunctionToTest"_when"StateToTest"_should"ExpectedBehavior".
Although we did not follow this convention completely, it may be helpful to refactor some test names to align with it.

contracts/governance/MentoToken.sol
baroooo: this is by solidity stye guide: https://docs.soliditylang.org/en/v0.8.19/style-guide.html#avoiding-naming-collisions
it avoids the name collision and also signals emission contract will be treated differently.

by the same convention, prefixed underscore should be used for non-external state vars and not params: https://docs.soliditylang.org/en/v0.8.19/style-guide.html#underscore-prefix-for-non-external-functions-and-variables

having said that, it is just a convention and we can stick to the old way

philbow61 (approved):
LGTM 👍

contracts/governance/MentoToken.sol
philbow61: seeing this i would say it looks good 👍

nvtaveras (approved):
🚢

PR #271, opened 2023-09-14
Feat: Emission contract

baroooo:
### Description

Implement the Emission Contract's that will mint and send the Mento tokens to a target depending on a predefined schedule.

Information about the formula can be found here: https://www.notion.so/Spike-Emission-contract-a3389802add2438ca482c6d18a607a2f

A 5th term (t^5 / 120A^5) is added to the formula to ensure the entire supply is minted.

### Other changes

Refactor common functionality in tests to a TestSetup contract

### Tested

Tested by unit and fuzz tests

### Related issues

- Fixes https://github.com/mento-protocol/mento-general/issues/306

### Backwards compatibility

N/A

### Documentation

N/A


contracts/governance/Emission.sol
```diff
@@ -0,0 +1,104 @@
...
+    mentoToken.mint(emissionTarget, amount);
+  }
+
+  /**
+   * @dev Calculate the releasable token amount using a predefined formula.
+   * The Maclaurin series is used to create a simpler approximation of the exponential decay formula.
+   * Original formula: E(t) = supply * exp(-A * t)
+   * Approximation: E(t) = supply * (1 - (t / A) + (t^2 / 2A^2) - (t^3 / 6A^3) + (t^4 / 24A^4))
+   * where A = HALF_LIFE / ln(2)
+   * @dev A 5th term (t^5 / 120A^5) is added to ensure the entire supply is minted around 31.5 years.
```
rcroessmann, line 80:
```suggestion
   * @dev A 5th term (t^5 / 120A^5) is added to ensure the entire supply is minted within around 31.5 years.
```

rcroessmann (approved):
Looks good from a math/economics perspective.

philbow61 (approved):
LGTM 👍 
Liked the usage of fuzz tests here 💯

test/governance/Emission.t.sol
```diff
@@ -0,0 +1,233 @@
+// solhint-disable func-name-mixedcase
+// SPDX-License-Identifier: GPL-3.0-or-later
+pragma solidity 0.8.18;
+
+import { TestSetup } from "./TestSetup.sol";
+
+contract EmissionTest is TestSetup {
+  uint256 public constant INITIAL_TREASURY_BALANCE = 100_000_000 * 1e18;
+  uint256 public constant NEGLIGIBLE_AMOUNT = 2e18;
```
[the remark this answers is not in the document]
baroooo: there will be difference of < 2 tokens between the calculated and emitted amount because of the precision lose. we decided 2 tokens is small enough to neglect. if the difference is larger, we should catch it with the tests

test/governance/TestSetup.sol
```diff
@@ -0,0 +1,34 @@
...
+import { Test } from "forge-std-next/Test.sol";
+
+import { MentoToken } from "contracts/governance/MentoToken.sol";
+import { Emission } from "contracts/governance/Emission.sol";
+
+contract TestSetup is Test {
+  MentoToken public mentoToken;
+  Emission public emission;
+
+  address public constant VESTING_CONTRACT = address(111);
```
bowd, line 14: Let's use [makeAddr](https://book.getfoundry.sh/reference/forge-std/make-addr) for these. It's nice that it also labels it automatically and you can see it in traces better.

bowd (approved):
Great work 👍, just one little nit around the test setup to make it nicer.

PR #274, opened 2023-09-17
Airgrab Contract

### LINKED ISSUE #282: Airgrab contract

bowd:
### Description

After a few iterations, we landed on a simple Airgrab scheme:
- With a list of <address>,<token amount> we generate a merkle tree and record the root in the contract 
- A user does KYC, gets the merkle proof from our backend and submits the claim
- The contract verifies the merkle proof and KYC and locks the tokens on behalf of the user for a predetermined cliff and slope.
- The airgrab has an endTimestamp after witch any leftover tokens in the contract can be transferred to the treasury.

More details on specifics around the KYC can be extracted from the comments in the code which contain links to documentation.

There is a single open question around how we encode the expected credential from fractal. The expected credential is basically the set of restrictions we want to enforce and it's a string that looks like `"level:plus;residency_not:ca,us"`.
I currently provided that statically in the code and it's probably not the final version that we want. But to parameterize in a naive way would incur a big gas because we can only store it in storage, and not as an immutable parameter because only immutable base types are allowed as of now in solidity. If we want to parameterize it the approach I would take would be to store a keccak of it as an immutable bytes32 parameter, and provide the full string as a parameter to the claim function and when verifying the credential we would first verify that the keccak matches and then use it in the construction of the full credential. I'm not sure if it's worth it or if it's fine to just come up with the value and keep it static in line as is right now.

### Other changes

N/A

### Tested

Yes

### Related issues

- Fixes #282 

### Backwards compatibility

N/A

### Documentation

N/A


contracts/governance/Airgrab.sol
```diff
@@ -0,0 +1,244 @@
...
+    emit TokensClaimed(account, amount, votingPower);
+  }
+
+  /**
+   * @dev Allows the treasury to reclaim any tokens after the airgrab has ended.
+   * @notice This function can only be called if the airgrab has ended.
+   * The function takes a token as a param in case the contract has been sent
+   * tokens other than the airgrab token.
+   */
+  function drain(address tokenToDrain) external nonReentrant {
```
philbow61, line 236: what is the scenario of draining a different token than the Mento token?

test/governance/Airgrab.t.sol
```diff
@@ -0,0 +1,469 @@
...
+  event TokensDrained(address indexed token, uint256 amount);
+
+  /// @notice setup for drain tests
+  modifier d_setUp() {
+    initAirgrab();
+    _;
+  }
+
+  /// @notice Reverts if airgrab hasn't ended
+  function test_Drain_beforeAirgrabEnds() public d_setUp {
```
philbow61, line 228: nit:
why are these tests marked as public while others on top are marked as external?

test/governance/Airgrab.t.sol
```diff
@@ -0,0 +1,469 @@
...
+  /// @notice Renounces ownership and sets token
+  function test_Initialize_TransfersOwnershipAndSetsToken() external i_setUp {
+    vm.expectEmit(true, true, true, true);
+    emit OwnershipTransferred(address(this), address(0));
+    vm.expectEmit(true, true, true, true);
+    emit Approval(address(airgrab), lockingContract, type(uint256).max);
+    i_subject();
+    assertEq(address(airgrab.token()), tokenAddress);
+    assertEq(airgrab.owner(), address(0));
+  }
```
philbow61, line 200: nit: 
can we add another assert to verify the allowance was set correctly

test/governance/Airgrab.t.sol
```diff
@@ -0,0 +1,469 @@
...
+    cl_subject();
+    vm.expectRevert("Airgrab: already claimed");
+    cl_subject();
+  }
+
+  // ========================================
+  // Airgrab.claim {claim}
+  // ========================================
+
+  /// @notice when the claimer has not enough balance
```
philbow61, line 443: should it be the Airgrab contract here?
`/// @notice when the Airgrab contract has not enough balance`

philbow61 (changes requested):
LGTM 🦈 
just some nits / questions

contracts/governance/Airgrab.sol
```diff
@@ -0,0 +1,244 @@
...
+   * @param merkleProof The merkle proof for the account.
+   */
+  modifier canClaim(
+    address account,
+    uint256 amount,
+    bytes32[] calldata merkleProof
+  ) {
+    require(block.timestamp <= endTimestamp, "Airgrab: finished");
+    require(!claimed[account], "Airgrab: already claimed");
+    bytes32 leaf = keccak256(bytes.concat(keccak256(abi.encode(account, uint256(amount)))));
```
baroooo, line 133: why do we do bytes.concat and second hash here ?

baroooo (changes requested):

contracts/governance/Airgrab.sol
```diff
@@ -0,0 +1,244 @@
...
+  )
+    external
+    hasValidKyc(account, fractalProof, fractalProofValidUntil, fractalProofApprovedAt, fractalId)
+    canClaim(account, amount, merkleProof)
+    nonReentrant
+  {
+    require(IERC20(token).balanceOf(address(this)) >= amount, "Airgrab: insufficient balance");
+
+    claimed[account] = true;
+    uint256 votingPower = lockingContract.lock(account, address(0), amount, slopePeriod, cliffPeriod);
```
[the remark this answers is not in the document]
bowd: Huh I dunno why I thought if your voting power is not delegated than it means you wield it, cool, I'll add delegate configuration at the contract level so we can set from the UI if we want to.

contracts/governance/Airgrab.sol
```diff
@@ -0,0 +1,244 @@
...
+   */
+  modifier hasValidKyc(
+    address account,
+    bytes memory proof,
+    uint256 validUntil,
+    uint256 approvedAt,
+    string memory fractalId
+  ) {
+    require(block.timestamp < validUntil, "Airgrab: KYC no longer valid");
+    require(fractalMaxAge == 0 || block.timestamp < approvedAt + fractalMaxAge, "Airgrab: KYC not recent enough");
```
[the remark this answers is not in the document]
bowd: This is how they suggest setting things up because I think their KYCs have a pretty long `validUntil` but it could be that our own business logic would require us to only accept kyc's with a certain freshness.

contracts/governance/Airgrab.sol
baroooo: oh ok, looks like it is a recent change made against second preimage attack:
https://flawed.net.nz/2018/02/21/attacking-merkle-trees-with-a-second-preimage-attack/

baroooo: I remember a great CTF challenge from paradigm which is related to this if anyone interested:

https://github.com/paradigmxyz/paradigm-ctf-2022/tree/main/merkledrop

contracts/governance/Airgrab.sol
```diff
@@ -0,0 +1,244 @@
...
+ * predetermined period.
+ * @dev The contract is only Ownable because of the circular dependency
+ * between Token and Airgrab. We use the initialize method to set the token address
+ * after the Token contract has been deployed, and renounce ownership.
+ */
+contract Airgrab is Ownable, ReentrancyGuard {
+  using SafeERC20 for IERC20;
+
+  uint256 public constant PRECISION = 1e18;
+  uint32 public constant MAX_CLIFF_PERIOD = 103;
```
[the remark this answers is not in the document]
bowd: So here's the logic:
4 years is 52 * 4 = 208 weeks
In the locking contract things are at week resolution so say time looks like this with week numbers in parentheses.
```
----(N)-------(N+1)-------(N+2)-------(N+4)--------
           ^ we want to lock here
```

If we select locking for one week, we will actually be snapping to (N) as the start date and have (N+2) as the end date. So, in effect we're locking for ~10 days as if the user started at (N).

So then, if we want the maximum to be 2 years, we actually want to select 2 years - 1 week so max of 207 week, which will effectively result in 208 weeks locking.

@baroooo correct me if I'm wrong here

contracts/governance/Airgrab.sol
```diff
@@ -0,0 +1,241 @@
...
+   */
+  function initialize(address token_) external onlyOwner {
+    require(token_ != address(0), "Airgrab: invalid token");
+    renounceOwnership();
+    token = IERC20(token_);
+    require(token.approve(address(lockingContract), type(uint256).max), "Airgrab: approval failed");
+  }
+
+  /**
+   * @dev Allows `account` to claim `amount` tokens if the merkle proof and kyc is valid.
```
philbow61, line 193: Can we update the comment because of the changes from `account -> msg.sender`

test/governance/Airgrab.t.sol
```diff
@@ -0,0 +1,490 @@
...
+  /// @notice when not in tree, it reverts
+  function test_Claim_canClaim_whenNotInTree_reverts() public cl_setUp {
+    cl_params.fractalProof = validKycSignature(fractalSignerPk, invalidClaimer, EXPECTED_CREDENTIAL);
+    cl_params.account = invalidClaimer;
+    vm.expectRevert("Airgrab: not in tree");
+    cl_subject();
+  }
+
+  /// @notice When the amount is not the right one for the claimer, it reverts
+  function test_Claim_canClaim_whenInvalidClaimAmount_reverts() public cl_setUp validKyc {
```
philbow61, line 417: Can we try to stick to our test naming convention here 😁

contracts/governance/Airgrab.sol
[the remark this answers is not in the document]
philbow61: ok i see 👍

philbow61 (approved):

contracts/governance/Airgrab.sol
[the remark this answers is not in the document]
baroooo: hmm I don't think so because it shifts everything back. my understanding is:
when you lock for a week between (N) and (N+1) the end date is (N+1). if you unlock 4 days later, you will receive full amount as 1 week is passed from contracts perspective, because 4 days later we are in a new epoch (N+1)

baroooo (approved):

.gitmodules
```diff
@@ -13,3 +13,6 @@
 [submodule "lib/forge-std-next"]
 	path = lib/forge-std-next
 	url = https://github.com/foundry-rs/forge-std
+[submodule "lib/locking-contracts"]
```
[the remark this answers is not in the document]
baroooo: I will update this once locking contracts are merged

test/governance/Airgrab.t.sol
baroooo: I think this is fine since it tests `canClaim` modifier on `claim` function

PR #278, opened 2023-09-25
Update fork tests to work with eXOF

bowd:
### Description

Fork tests weren't passing on `baklava` after the eXOF update because:
1. We have an unrecoverable breaker
2. We have trading limits on both assets in a pool
3. We have multiple dependencies on a single rate

The approaches I took to fix were:
1. Include manual reset logic to our helper functions that used to reset breakers just by altering the median
2. Check if both assets have limits and in that case choose to test the limit that is more restrictive
3. Slightly alter the logic for rate feed dependencies to test when there are multiple dependencies for a rate feed.

For (3) I initially tried an approach where I was using low-level `call` to figure out how many dependencies a rate feed has but that threw me into a rabbit hole of weird foundry behaviour because catching those `call` reverts ended up consuming a lot of gas and finally the function ended with out-of-gas errors. This was in an attempt to remove the hardcoded elements from the test, but I think we need to live with them for now.

### Other changes

N/A

### Tested

Test run

### Related issues

- Fixes https://github.com/mento-protocol/mento-general/issues/302

### Backwards compatibility

Yes

### Documentation

N/A

┆Issue is synchronized with this [Jira Story](https://mentolabs.atlassian.net/browse/MC-12) by [Unito](https://www.unito.io)

### DIFF
--- a/test/fork-tests/BaseForkTest.t.sol
+++ b/test/fork-tests/BaseForkTest.t.sol
@@ -115,7 +115,6 @@ contract BaseForkTest is Test, TokenHelpers, TestAsserts {
     for (uint256 i = 0; i < exchangeProviders.length; i++) {
       IExchangeProvider.Exchange[] memory _exchanges = IExchangeProvider(exchangeProviders[i]).getExchanges();
       for (uint256 j = 0; j < _exchanges.length; j++) {
-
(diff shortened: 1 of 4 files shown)


nvtaveras:
@bowd made some cleanups, mostly removed debugging logs and unused code

nvtaveras (approved):

PR #302, opened 2023-11-22
fix: add null addresses checks in creditGasFees

nvtaveras:
### Description

While testing MU04 on Baklava we noticed that gas payment with stables was broken and after some digging we realized that this is due to a slight change in the implementation details of the `creditGasFees` function after switching to the StableTokenV2 implementation. 

TL;DR is that there were checks for the null address in the v1 implementation that were removed in v2, which makes the tx fail when the node attempts to include it in a block due to the `gatewayFeeRecipient` address being `0x` and the underlying call to `_transfer(from, 0x, amount)` failing. 

This change adds a null address check for each recipient addresses and it also keeps track of any amount that can't be transferred to recipients so that it's burned at the end of the function for proper accounting purposes.

### Other changes

Fixed a slight typo of `conmunityFund` which was repeated throughout the code.

### Tested

- Wrote unit tests passing null addresses for each recipient
- Existing tests pass as expected

### Backwards compatibility

This is actually backwards compatible with the original V1 implementation that has proper null checks before the balances of the different accounts are manipulated https://github.com/mento-protocol/mento-core/blob/develop/contracts/legacy/StableToken.sol#L581-L587

### DIFF
--- a/contracts/tokens/StableTokenV2.sol
+++ b/contracts/tokens/StableTokenV2.sol
@@ -291,9 +291,29 @@ contract StableTokenV2 is ERC20PermitUpgradeable, IStableTokenV2, CalledByVm {
     uint256 gatewayFee,
     uint256 baseTxFee
   ) external onlyVm {
+    uint256 amountToBurn = 0;
     _mint(from, refund + tipTxFee + gatewayFee + baseTxFee);
-    _transfer(from, feeRecipient, tipTxFee);
-    _transfer(from, gatewayFeeRecipient, gatewayFee);
-    _transfer(from, communityFund, baseTxFee);
+
+    if (feeRecipient != address(0)) {
+      _transfer(from, feeRecipient, tipTxFee);
+    } else if (tipTxFee > 0) {
+      amountToBurn += tipTxFee;
+    }
+
+    if (gatewayFeeRecipient != address(0)) {
+      _transfer(from, gatewayFeeRecipient, gatewayFee);
+    } else if (gatewayFee > 0) {
+      amountToBurn += gatewayFee;
+    }
+
+    if (communityFund != address(0)) {
+      _transfer(from, communityFund, baseTxFee);
+    } else if (baseTxFee > 0) {
+      amountToBurn += baseTxFee;
+    }
+
+    if (amountToBurn > 0) {
+      _burn(from, amountToBurn);
+    }
   }
 }
--- a/contracts/tokens/StableTokenV2.sol
+++ b/contracts/tokens/StableTokenV2.sol
@@ -258,8 +258,7 @@ contract StableTokenV2 is ERC20PermitUpgradeable, IStableTokenV2, CalledByVm {
    * @param value The amount of balance to reserve
    * @dev Note that this function is called by the protocol when paying for tx fees in this
    * currency. After the tx is executed, gas is refunded to the sender and credited to the
-   * various tx fee recipients via a call to `creditGasFees`. Note too that the events emitted
-   * by `creditGasFees` reflect the *net* gas fee payments for the transaction.
+   * various tx fee recipients via a call to `creditGasFees`.
    */
   function debitGasFees(address from, uint256 value) external onlyVm {
     _burn(from, value);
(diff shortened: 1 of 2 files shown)


contracts/tokens/StableTokenV2.sol
```diff
@@ -291,9 +291,29 @@ contract StableTokenV2 is ERC20PermitUpgradeable, IStableTokenV2, CalledByVm {
     uint256 gatewayFee,
     uint256 baseTxFee
   ) external onlyVm {
+    uint256 amountToBurn = 0;
```
baroooo, line 294: uint is 0 by default
```suggestion
    uint256 amountToBurn;
```

contracts/tokens/StableTokenV2.sol
```diff
@@ -291,9 +291,29 @@ contract StableTokenV2 is ERC20PermitUpgradeable, IStableTokenV2, CalledByVm {
     uint256 gatewayFee,
```
baroooo, line 291:
```
 Note too that the events emitted by `creditGasFees` reflect the *net* gas fee
   * payments for the transaction.
```

I dont think the comment above is correct anymore. We now have extra transfer events emitted for mints and burns. Amounts will cancel eachother but the number of events emitted wont be same as the legacy implementation, should be fine though, assuming they are not used for any other reason.

test/tokens/StableTokenV2.t.sol
```diff
@@ -187,7 +188,103 @@ contract StableTokenV2Test is BaseTest {
...
+    uint256 baseTxFee = 40;
+    uint256 holder0InitialBalance = token.balanceOf(holder0);
+
+    vm.prank(address(0));
+    token.creditGasFees(holder0, feeRecipient, address(0), communityFund, refund, tipTxFee, gatewayFee, baseTxFee);
+
+    assertEq(token.balanceOf(holder0), holder0InitialBalance + refund);
+    assertEq(token.balanceOf(feeRecipient), tipTxFee);
+    assertEq(token.balanceOf(gatewayFeeRecipient), 0);
+    assertEq(token.balanceOf(communityFund), baseTxFee);
```
baroooo, line 232: can we also verify if the totalSupply is updated correctly?

contracts/tokens/StableTokenV2.sol
```diff
@@ -291,9 +291,29 @@ contract StableTokenV2 is ERC20PermitUpgradeable, IStableTokenV2, CalledByVm {
...
     _mint(from, refund + tipTxFee + gatewayFee + baseTxFee);
-    _transfer(from, feeRecipient, tipTxFee);
-    _transfer(from, gatewayFeeRecipient, gatewayFee);
-    _transfer(from, communityFund, baseTxFee);
+
+    if (feeRecipient != address(0)) {
+      _transfer(from, feeRecipient, tipTxFee);
+    } else if (tipTxFee > 0) {
+      amountToBurn += tipTxFee;
+    }
```
baroooo, line 301: wdyt about moving this duplicate logic to an internal function?

contracts/tokens/StableTokenV2.sol
```diff
@@ -291,9 +291,29 @@ contract StableTokenV2 is ERC20PermitUpgradeable, IStableTokenV2, CalledByVm {
...
+
+    if (gatewayFeeRecipient != address(0)) {
+      _transfer(from, gatewayFeeRecipient, gatewayFee);
+    } else if (gatewayFee > 0) {
```
philbow61, line 305: Do we need those checks for the values being larger than 0 here?
If not would be more gas-efficient to remove them?

test/tokens/StableTokenV2.t.sol
nvtaveras: good point, will add to all tests

contracts/tokens/StableTokenV2.sol
nvtaveras: Hmm I think I'd be fine with keeping as is as it's very specific to this function and having the explicit checks each time doesn't make this code feel messy (imho).

contracts/tokens/StableTokenV2.sol
nvtaveras: I ran some gas tests with Foundry and in the most common case where `gatewayFee == 0` and `gatewayFeeRecipient == address(0)` it's a bit cheaper with checks (76879 with vs 76927 without). The case when there is a burn it's actually a bit more expensive (79375 with checks vs 79341 without), so seems like we can leave it as it is

contracts/tokens/StableTokenV2.sol
nvtaveras: Hmm yeah, I also think that explaining some of the logic here is a bit tricky since it's technically tied to the node implementation. What do you think about just removing that comment?

baroooo: sure, sounds good 👍

baroooo (approved):

philbow61 (approved):
LGTM ⛷️

PR #315, opened 2023-12-01
fix: add null addresses checks in creditGasFees (#302)

nvtaveras:
### Description

This merges the `StableTokenV2` contract fix for using stable tokens as gas currencies. This was done on top of the `2.2.0` tag and reviewed by Macro.

### Other changes

-


philbow61 (approved):

nvtaveras:
Need to merge a slither fix which can't be done in this PR due to the branch, so will do it in https://github.com/mento-protocol/mento-core/pull/316 instead

PR #367, opened 2024-02-12
feat: Improve how the initial allocation is forwarded.

bowd:
### Description

When we moved to a dynamic initial allocation calculation, we did something that has the potential for confusion: we're passing in two arrays to the GovernanceFactory contract:
- additionalAllocationRecipients
- allocationAmounts
With the implicit idea that `allocationAmounts.length == additionalAllocationRecipients.length + 2` and that the first two `allocationAmounts` relate to the airgrab and the Mento Treasury.

Even tho this is a one time thing I find that this adds the potential for confusion and will also make reviews more complicated.
I propose simplifying this by introducing an initial allocation struct:
```solidity
  struct MentoTokenAllocationParams {
    uint256 airgrabAllocation;
    uint256 mentoTreasuryAllocation;
    address[] additionalAllocationRecipients;
    uint256[] additionalAllocationAmounts;
  }
```

This explicitly defines the two "implicit" allocations and then has an array of 1:1 additional recipients and amounts.
It's also nicer to read when reviewing the CGP (in the cgp-viewer app I wrote): 
image

### Other changes

N/A

### Tested

- Tests covered

### Backwards compatibility

N/A

### Documentation

N/A

### DIFF
--- a/contracts/governance/GovernanceFactory.sol
+++ b/contracts/governance/GovernanceFactory.sol
@@ -41,6 +41,14 @@ contract GovernanceFactory is Ownable {
     address mentoGovernor
   );
 
+  /// @dev Parameters for the initial token allocation
+  struct MentoTokenAllocationParams {
+    uint256 airgrabAllocation;
+    uint256 mentoTreasuryAllocation;
+    address[] additionalAllocationRecipients;
+    uint256[] additionalAllocationAmounts;
+  }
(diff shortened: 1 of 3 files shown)


chapati23 (approved):
lgtm 👍

baroooo (approved):
✅

PR #371, opened 2024-02-14
feat: make ci ✅ with 💚 and 🥑s

bowd:
### Description

- Split jobs for better readability
- Make storage checks run only on push to develop (i.e. not for each commit of a work in progress branch) in an attempt to not hit rate limits.
- removed the slither.db.json and inlined ignores - it's important to actually review these and see if we want to actually fix any of them
- made slither ignore contracts/common and contracts/legacy . Common will be moved to @celo/contracts npm dependency anyway.
- fixed echidna compilation issue, yey!

### Other changes

- Moved `SortedOracles.sol` to `common` because moving forward that contract will be owned by `clabs` and we will need to deploy it through their pipeline. So everything in `common` will move to `@celo/contracts`, including SortedOracles.

### Tested

In progress

### Related issues

N/A

### Backwards compatibility

N/A

### Documentation

N/A


.github/workflows/lint_test.yaml
```diff
@@ -0,0 +1,64 @@
...
+    steps:
+      - name: "Check out the repo"
+        uses: "actions/checkout@v3"
+        with:
+          submodules: "recursive"
+
+      - name: "Install Foundry"
+        uses: "onbjerg/foundry-toolchain@v1"
+        with:
+          version: "nightly-d369d2486f85576eec4ca41d277391dfdae21ba7"
```
chapati23, line 32: while we're at it, should we update to latest foundry release:
[nightly-fd87629fbc4ae2e0fa00ccf42b4a9ebe1b521d55](https://github.com/foundry-rs/foundry/releases/tag/nightly-fd87629fbc4ae2e0fa00ccf42b4a9ebe1b521d55)

```suggestion
          version: "nightly-fd87629fbc4ae2e0fa00ccf42b4a9ebe1b521d55"
```

.github/workflows/echidna.yaml
```diff
@@ -0,0 +1,58 @@
...
+      matrix:
+        contract: ${{ fromJson(needs.get_echidna_test_names.outputs.matrix) }}
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          submodules: recursive
+      - name: Install Foundry
+        uses: onbjerg/foundry-toolchain@v1
+        with:
+          version: "nightly-d369d2486f85576eec4ca41d277391dfdae21ba7"
```
chapati23, line 47: while we're at it, should we update to latest foundry release:
[nightly-fd87629fbc4ae2e0fa00ccf42b4a9ebe1b521d55](https://github.com/foundry-rs/foundry/releases/tag/nightly-fd87629fbc4ae2e0fa00ccf42b4a9ebe1b521d55)

```suggestion
          version: "nightly-fd87629fbc4ae2e0fa00ccf42b4a9ebe1b521d55"
```

.github/workflows/lint_test.yaml
```diff
@@ -0,0 +1,64 @@
...
+
+      - name: "Install Node.js"
+        uses: "actions/setup-node@v3"
+        with:
+          cache: "yarn"
+          node-version: "16"
```
chapati23, line 38: node 16 [reached end-of-life in october '23](https://nodejs.org/en/about/previous-releases), maybe good time to try if latest LTS (20) works?

```suggestion
          node-version: "20"
```

.github/workflows/slither.yaml
```diff
@@ -0,0 +1,53 @@
...
+    runs-on: "ubuntu-latest"
+    steps:
+      - name: "Check out the repo"
+        uses: "actions/checkout@v3"
+        with:
+          submodules: "recursive"
+      - name: "Install Foundry"
+        uses: "onbjerg/foundry-toolchain@v1"
+        with:
+          version: "nightly-d369d2486f85576eec4ca41d277391dfdae21ba7"
```
chapati23, line 31: copy pasta from above 

```suggestion
          version: "nightly-fd87629fbc4ae2e0fa00ccf42b4a9ebe1b521d55"
```

.github/workflows/slither.yaml
```diff
@@ -0,0 +1,53 @@
...
+      - name: "Build the contracts"
+        run: |
+          forge build --skip test --skip script --build-info
```
chapati23, line 34: i'm not sure this extra build step is still needed. think slither got an update at some point to fix this.

if I do this locally:

- `forge clean`
- `slither .`

it works fine and slither is smart enough to clean and build on its own

image

.github/workflows/storage-layout.yaml
```diff
@@ -0,0 +1,38 @@
...
+          - contracts/oracles/SortedOracles.sol:SortedOracles
+          - contracts/tokens/StableTokenV2.sol:StableTokenV2
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          submodules: recursive
+      - name: Install Foundry
+        uses: onbjerg/foundry-toolchain@v1
+        with:
+          version: "nightly-d369d2486f85576eec4ca41d277391dfdae21ba7"
```
chapati23, line 34: copy pasta from above 

```suggestion
          version: "nightly-fd87629fbc4ae2e0fa00ccf42b4a9ebe1b521d55"
```

slither.config.json
```diff
@@ -0,0 +1,5 @@
+{
+  "foundry_ignore_compile": true,
```
chapati23, line 2: locally for me it looks fine to let slither run foundry, what's the reason we need this?

chapati23 (commented):
CI is greeeeeeeen 🤩🤩🤩

lgtm, just some questions/suggestions

.github/workflows/slither.yaml
bowd: Ah nice! Will test this out!

bowd:
@philipraetsch thanks for the suggestions!
I've updated the foundry version and removed the build steps from slither.
We still need the build step in echidna because they haven't update [echidna-action](https://github.com/crytic/echidna-action) to install Foundry in the docker. They did update [slither-action](https://github.com/crytic/slither-action) tho' so if anybody wants to be a good samaritan you can open a PR there :)

chapati23 (approved):

contracts/swap/Reserve.sol
```diff
@@ -549,6 +556,7 @@ contract Reserve is IReserve, ICeloVersionedContract, Ownable, Initializable, Us
    */
   function getOrComputeTobinTax() external nonReentrant returns (uint256, uint256) {
     // solhint-disable-next-line not-rely-on-time
```
baroooo, line 558: is this the old name for timestamp matcher?

baroooo (approved):
I believe we can disable `timestamp` matcher at the config level. Because,
1- I don't think we rely on it at a critical level
2- Not sure about specifics of Celo, but it is much harder to exploit in a PoS setup
3- Timestamps are used frequently in our repo

The rest looks good as it is.

PR #374, opened 2024-02-16
chore: fixed solhint warnings

chapati23:
### What
Fixed relevant and disabled irrelevant solhint warnings

### Review
Check solhint passes without any warnings now


test/utils/BaseTest.next.sol
```diff
@@ -13,15 +13,15 @@ contract BaseTest is Test {
   address public constant REGISTRY_ADDRESS = 0x000000000000000000000000000000000000ce10;
   IRegistry public registry = IRegistry(REGISTRY_ADDRESS);
 
-  address public constant deployer = address(0x31337);
+  address public constant DEPLOYER = address(0x31337);
```
bowd, line 16: There are _a lot_  of references to this `deployer` in tests extending both this and the `0.5` version. I would say we maybe ignore the solhint warning here or make it a storage variable instead of a constant. Or if you're really feeling it, go deep and make all references uppercase :)

bowd (changes requested):
Looking good 🤩 just need to fix the `deployer -> DEPLOYER` situation to get CI green.

bowd (approved):

contracts/governance/TimelockController.sol
```diff
@@ -2,7 +2,9 @@
 pragma solidity 0.8.18;
 
 // solhint-disable-next-line max-line-length
```
nvtaveras, line 4: no longer need this after the change?

contracts/governance/deployers/ProxyDeployerLib.sol
```diff
@@ -2,7 +2,9 @@
 pragma solidity 0.8.18;
 
 //solhint-disable-next-line max-line-length
```
nvtaveras, line 4: ditto

contracts/tokens/patched/ERC20Upgradeable.sol
```diff
@@ -52,6 +53,8 @@ contract ERC20Upgradeable is Ownable, Initializable, IERC20Upgradeable, IERC20Me
   uint256[4] private __deprecated_inflationState_storage_slot__;
   bytes32 private __deprecated_exchangeRegistryId_storage_slot__;
 
+  // solhint-enable var-name-mixedcase
+
```
nvtaveras, line 57: delete?

test/fork-tests/TokenUpgrade.t.sol
```diff
@@ -12,6 +12,7 @@ contract TokenUpgradeForkTest is Test {
   address public constant REGISTRY_ADDRESS = 0x000000000000000000000000000000000000ce10;
   IRegistry public registry = IRegistry(REGISTRY_ADDRESS);
 
+  // solhint-disable-next-line func-name-mixedcase
```
nvtaveras, line 15: is the mixedcase here because of the `_` ?

nvtaveras (changes requested):

contracts/governance/TimelockController.sol
chapati23: still needed unfortunately, if i remove it it complains

contracts/tokens/patched/ERC20Upgradeable.sol
chapati23: this is re-enabling the rule after it was disabled for the previous block, it's a bit cleaner IMO because at least it's then linting the local var definitions instead of completely ignoring this rule

test/fork-tests/TokenUpgrade.t.sol
chapati23: yes

nvtaveras (approved):
Thanks for answering the comments!

PR #375, opened 2024-02-19
Update locking formula

### LINKED ISSUE #373: Modify locking voting power formula
**Description** 

Currently the voting power formula looks like this (slightly simplified):

```math
P(t,s,c) = t * (0.2 + 0.8 * c / c_{max} + 0.6 * s / s_{max})
```

This has the following properties:
1. Locking for the minimum times still gives you 20% of your token as voting power.
2. Locking for the full cliff gives you 100% of the voting power.
3. Adding a slope boosts the voting power beyond 100%.

After discussion, we have decided that we don't see the utility of some of these properties. So, we want to bring the system closer to the classical veCurve formula, by changing it to:

```math
P(t,s,c) = t * min(s/s_{max} + c/c_{max}, 1)
```

> In reality, we need to scale both `cliff` and `slope` between a minimum value and a maximum. $$s/s_{max}$$ should be replaced with $$(s - s_{min}) / (s_{max} - s_{min})$$, it is equivalent to $$s/s_{max}$$ when $$s_{min} = 0$$ I used the simplified version not to complicate the intent behind the formula.

This gives us that you can achieve 100% by maxing out either `slope` or `cliff` individually or a linear combination of the two. 
The formula will ensure that the final multiplier is capped at 100% and has the downside of allowing for disadvantageous scenarios. We could consider reverting if the calculated multiplier is > 1 and thus forcing the UI to choose only advantageous scenarios. But this might bring issues because of rounding ( i.e. `P(t,s',c') = 0.95 and P(t, s'+1,c') = 1.05 and P(t, s', c'+1) = 1.02` ). 

We also need to research a bit more into whether this change will affect any assumptions made in the implementation.

**Acceptance criteria**

- [ ] Formula is implemented and thoroughly tested against future scenarios.
- [ ] Enough time is spent understanding what effect it could have on the code.

bowd:
### Description

Currently the voting power formula looks like this (slightly simplified):

```math
P(t,s,c) = t * (0.2 + 0.8 * c / c_{max} + 0.6 * s / s_{max})
```

We want to bring the system closer to the classical veCurve formula, by changing it to:

```math
P(t,s,c) = t * min(s/s_{max} + c/c_{max}, 1)
```

More details can be seen here https://github.com/mento-protocol/mento-core/issues/373

Since this formula sits in the core of the voting power calculations, reviews should be on a broader level. Please go through code changes, test calculations and more importantly how this may affect the whole locking mechanism.

Known issues:
- users can lock with a suboptimal schedule if the summed multiplier goes above 1. 

### Other changes

Updated the tests to use the exact amounts and not approximations
Removed unused imports

### Tested

Fixed the tests related to change. Actual calculations are added as comments.
Some extra fuzz tests are WIP and will be added with a separate PR

### Related issues

- Fixes #373

### DIFF
--- a/contracts/governance/locking/LockingBase.sol
+++ b/contracts/governance/locking/LockingBase.sol
@@ -223,10 +223,8 @@ abstract contract LockingBase is OwnableUpgradeable, IVotesUpgradeable {
     require(cliff >= minCliffPeriod, "cliff period < minimal lock period");
     require(slopePeriod >= minSlopePeriod, "slope period < minimal lock period");
 
-    uint96 cliffSide = (uint96(cliff - uint32(minCliffPeriod)) * (ST_FORMULA_CLIFF_MULTIPLIER)) /
-      (MAX_CLIFF_PERIOD - uint32(minCliffPeriod));
-    uint96 slopeSide = (uint96((slopePeriod - uint32(minSlopePeriod))) * (ST_FORMULA_SLOPE_MULTIPLIER)) /
-      (MAX_SLOPE_PERIOD - uint32(minSlopePeriod));
+    uint96 cliffSide = (uint96(cliff) * ST_FORMULA_CLIFF_MULTIPLIER) / MAX_CLIFF_PERIOD;
+    uint96 slopeSide = (uint96(slopePeriod) * ST_FORMULA_SLOPE_MULTIPLIER) / MAX_SLOPE_PERIOD;
     uint96 multiplier = cliffSide + slopeSide;
 
     if (multiplier > ST_FORMULA_DIVIDER) {
--- a/contracts/governance/locking/LockingBase.sol
+++ b/contracts/governance/locking/LockingBase.sol
@@ -190,9 +190,7 @@ abstract contract LockingBase is OwnableUpgradeable, IVotesUpgradeable {
 
   /**
    * @dev Сalculate and return (newAmount, newSlope), using formula:
-   * P = t * min(scale(c, c_min, c_max) + scale(s, s_min, s_max), 1),
-   * where:
-   *    scale(v, min, max) = (v - min) / (max - min)
+   * P = t * min(c/c_max + s/s_max, 1),
    *
    * The formula has the following properties:
    * - the voting power can't exceed the amount of tokens locked.
@@ -208,8 +206,8 @@ abstract contract LockingBase is OwnableUpgradeable, IVotesUpgradeable {
    * locking = (
    *   tokens *
    *   min(
-   *    (ST_FORMULA_CLIFF_MULTIPLIER * (cliffPeriod - minCliffPeriod))/(MAX_CLIFF_PERIOD - minCliffPeriod) +
-   *    (ST_FORMULA_SLOPE_MULTIPLIER * (slopePeriod - minSlopePeriod))/(MAX_SLOPE_PERIOD - minSlopePeriod),
+   *    (ST_FORMULA_CLIFF_MULTIPLIER * cliffPeriod) / MAX_CLIFF_PERIOD +
+   *    (ST_FORMULA_SLOPE_MULTIPLIER * slopePeriod) / MAX_SLOPE_PERIOD,
    *    ST_FORMULA_DIVIDER
    *   )
    * ) / ST_FORMULA_DIVIDER
--- a/contracts/governance/locking/LockingBase.sol
+++ b/contracts/governance/locking/LockingBase.sol
@@ -186,7 +186,7 @@ abstract contract LockingBase is OwnableUpgradeable, IVotesUpgradeable {
   }
 
   /**
-   * @dev Сalculate and return (newAmount, newSlope), using formula:
+   * @dev Сalculate and return (lockAmount, lockSlope), using formula:
    * P = t * min(c/c_max + s/s_max, 1),
    *
    * The formula has the following properties:
@@ -199,8 +199,7 @@ abstract contract LockingBase is OwnableUpgradeable, IVotesUpgradeable {
    *
    *
    * The formula roughly translates to solidity as:
-   * ```
-   * locking = (
+   * votingPower = (
    *   tokens *
    *   min(
    *    (ST_FORMULA_BASIS * cliffPeriod) / MAX_CLIFF_PERIOD +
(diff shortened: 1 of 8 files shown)


contracts/governance/locking/LockingBase.sol
```diff
@@ -18,10 +18,9 @@ abstract contract LockingBase is OwnableUpgradeable, IVotesUpgradeable {
   uint32 constant MAX_CLIFF_PERIOD = 103;
   uint32 constant MAX_SLOPE_PERIOD = 104;
 
-  uint32 constant ST_FORMULA_DIVIDER = 1 * (10**8); //stFormula divider          100000000
-  uint32 constant ST_FORMULA_CONST_MULTIPLIER = 2 * (10**7); //stFormula const multiplier  20000000
-  uint32 constant ST_FORMULA_CLIFF_MULTIPLIER = 8 * (10**7); //stFormula cliff multiplier  80000000
-  uint32 constant ST_FORMULA_SLOPE_MULTIPLIER = 4 * (10**7); //stFormula slope multiplier  40000000
+  uint32 constant ST_FORMULA_DIVIDER = 1 * (10**8); //stFormula divider          100000000 / 1.0 / 100%
```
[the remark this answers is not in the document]
chapati23: reading through the code a bit more, a name like `MAX_MULTIPLIER` would make more sense to me, but not feeling too strongly about, just a little nit

contracts/governance/locking/LockingBase.sol
```diff
@@ -18,10 +18,9 @@ abstract contract LockingBase is OwnableUpgradeable, IVotesUpgradeable {
...
+  uint32 constant ST_FORMULA_CLIFF_MULTIPLIER = 1 * (10**8); //stFormula cliff multiplier  100000000 / 1.0 / 100%
```
chapati23, line 22: for transparency to other readers here, we discussed on discord to potentially completely remove `ST_FORMULA_CLIFF_MULTIPLIER` and `ST_FORMULA_SLOPE_MULTIPLIER` as they're currently just set to 1 and not really doing anything other than making the code a bit more complex

given deadline is tight, decision is for @baroooo to make. no great harm in leaving it in as is either

contracts/governance/locking/LockingBase.sol
```diff
@@ -190,12 +189,29 @@ abstract contract LockingBase is OwnableUpgradeable, IVotesUpgradeable {
...
+   * The formula has the following properties:
+   * - the voting power can't exceed the amount of tokens locked.
+   * - a voter can reach 100% voting power by relying on either the slope or the cliff,
+   *   or a combination of both.
+   * - there is a parameter space above a diagonal on the (c, s) plane where the
+   *   voting power is capped at 100%, moving past that diagonal is disadvantageous
+   *   but the contract doesn't forbid it.
+   *
+   *
+   * The formulare roughly translates to solidity as:
```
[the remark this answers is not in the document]
baroooo: lol, looks like someone was filling some german formulare recently :D

contracts/governance/locking/LockingBase.sol
baroooo: Removed them 👍

chapati23 (approved):
🚀

contracts/governance/locking/LockingBase.sol
```diff
@@ -190,12 +186,29 @@ abstract contract LockingBase is OwnableUpgradeable, IVotesUpgradeable {
   }
 
   /**
-   * Сalculate and return (newAmount, newSlope), using formula:
-   * locking = (tokens * (
-   *      ST_FORMULA_CONST_MULTIPLIER
-   *      + ST_FORMULA_CLIFF_MULTIPLIER * (cliffPeriod - minCliffPeriod))/(MAX_CLIFF_PERIOD - minCliffPeriod)
-   *      + ST_FORMULA_SLOPE_MULTIPLIER * (slopePeriod - minSlopePeriod))/(MAX_SLOPE_PERIOD - minSlopePeriod)
-   *      )) / ST_FORMULA_DIVIDER
+   * @dev Сalculate and return (newAmount, newSlope), using formula:
```
nvtaveras, line 189: lockAmount, lockSlope?

contracts/governance/locking/LockingBase.sol
```diff
@@ -190,12 +186,29 @@ abstract contract LockingBase is OwnableUpgradeable, IVotesUpgradeable {
...
+   * - a voter can reach 100% voting power by relying on either the slope or the cliff,
+   *   or a combination of both.
+   * - there is a parameter space above a diagonal on the (c, s) plane where the
+   *   voting power is capped at 100%, moving past that diagonal is disadvantageous
+   *   but the contract doesn't forbid it.
+   *
+   *
+   * The formula roughly translates to solidity as:
+   * ```
+   * locking = (
```
nvtaveras, line 203: do we maybe want to rename this to something closer to the P in the formula? `power` | `votingPower` ?

test/governance/Locking/locking.t.sol
```diff
@@ -213,29 +214,37 @@ contract Lock_Locking_Test is Locking_Test {
...
     amount = 104000;
-    slopePeriod = 1;
+    slopePeriod = 24;
     cliff = 103;
     (lockAmount, lockSlope) = locking.getLock(amount, slopePeriod, cliff);
-    assertEq(lockAmount, 104399);
-    assertEq(lockSlope, 104399);
+    // floor ((24 / 104 + 103 / 103) * 104000) = 128000 > 104000
+    assertEq(lockAmount, 104000);
+    // divUp (104000, 1) = 105000
```
nvtaveras, line 246:
```suggestion
    // divUp (104000, 24) = 4334
```

nvtaveras (approved):

contracts/governance/locking/LockingBase.sol
```diff
@@ -205,14 +216,17 @@ abstract contract LockingBase is OwnableUpgradeable, IVotesUpgradeable {
...
+    uint96 multiplier = cliffSide + slopeSide;
+
+    if (multiplier > ST_FORMULA_BASIS) {
+      multiplier = ST_FORMULA_BASIS;
+    }
 
     uint256 amountMultiplied = uint256(amount) * uint256(multiplier);
-    lockAmount = uint96(amountMultiplied / (ST_FORMULA_DIVIDER));
+    lockAmount = uint96(amountMultiplied / (ST_FORMULA_BASIS));
+    require(lockAmount > 0, "voting power is 0");
```
bowd, line 229: Not sure if this `require` is really needed anymore unless both minCliffPeriod and minSlopePeriod are configured to 0, but it's probably good to have anyway

Issue #377, opened 2024-02-29
[Audit Finding] Mento tokens can be lost, or delegated incorrectly in `lock()` function


chapati23:
### Description 
Address audit finding: https://www.notion.so/0xmacro/Mento-3-Preliminary-Audit-Report-external-private-06ee6ce11d7d4df0a950ca97c75a2766?pvs=4#21daa0832c364978a2227fb05fba34f2

- [x] add require checks for both delegate and account
- [x] try to reach out to rari find out any potential use cases for using address(0)
Pinged them, waiting for an answer
- [x] check their audit to fid out whether this came up in theirs as well
[The report](https://chainsecurity.com/wp-content/uploads/2021/10/ChainSecurity_Rarible_Inc_Staking_and_Tokens_Smart-Contract-Audit_July-14th-2021.pdf) does not have a finding related to the issue. 
- [x] Address finding

PR #392, opened 2024-03-05
fix: check for locker and delegate addresses

baroooo:
### Description

fixes[[M-1] Mento tokens can be lost, or delegated incorrectly in `lock()` function](https://www.notion.so/0xmacro/Mento-3-Preliminary-Audit-Report-external-private-06ee6ce11d7d4df0a950ca97c75a2766?pvs=4#21daa0832c364978a2227fb05fba34f2)

- added zero address check for locker and delegate in lock() as suggested
- also added checks for delegates in delegateTo() and relock() because the same issue is present in those functions

### Other changes
No

### Tested

- New unit tests implemented
- Old tests pass

### Related issues

- Fixes [#377](https://github.com/mento-protocol/mento-core/issues/377)


denviljclarke (approved):
LGTM

PR #396, opened 2024-03-11
fix: hardcoded emission supply

### LINKED ISSUE #395: Emission Supply Bug

baroooo:
### Description

We forgot to update hardcoded emission supply in the emission contract after making allocations parametrised. This may cause issues with the emission calculation when the emission supply in the token contract is not equal to the hardcoded emission supply in the emission contract.

This fix removes the hardcoded emission supply, and sets it through init call from the governance factory.

### Other changes

No

### Tested

Fixed unit tests

### Related issues

- Fixes #395

### DIFF
--- a/contracts/governance/GovernanceFactory.sol
+++ b/contracts/governance/GovernanceFactory.sol
@@ -116,8 +116,8 @@ contract GovernanceFactory is Ownable {
     celoCommunityFund = celoCommunityFund_;
 
     // Precalculated contract addresses:
-    address emissionPrecalculated = addressForNonce(3);
-    address tokenPrecalculated = addressForNonce(4);
+    address tokenPrecalculated = addressForNonce(2);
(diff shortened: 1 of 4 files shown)


contracts/governance/GovernanceFactory.sol
```diff
@@ -116,8 +116,8 @@ contract GovernanceFactory is Ownable {
     celoCommunityFund = celoCommunityFund_;
 
     // Precalculated contract addresses:
-    address emissionPrecalculated = addressForNonce(3);
-    address tokenPrecalculated = addressForNonce(4);
+    address tokenPrecalculated = addressForNonce(2);
+    address emissionPrecalculated = addressForNonce(4);
```
bayological, line 120: Why is this not 3?

bayological (approved):

philbow61 (approved):

PR #412, opened 2024-04-12
Fix/sherlock finding nr 4

### LINKED ISSUE #407: Fix contest finding #4
### Description

https://github.com/sherlock-audit/2024-02-mento-judging/issues/4

We decided to fix it by passing the history as a storage variable.

We should also add tests to measure the gas cost of the functions that use balanceOf and totalSupply with a high number of locks and various timings such as with a new lock just before a proposal creation, after a proposal creation, in the same block with a castVote etc.

### Acceptance Criteria
- [x] Issue fixed
- [x] Gas cost is reasonable
- [x] PR is opened to fixes branch
- [x] Fix is added to the sherlock fixes review

baroooo:
### Description

This PR uses history storage var instead of memory variable to fix [Sherlock contest finding #4](https://github.com/sherlock-audit/2024-02-mento-judging/issues/4)
It also introduces a lower limit of 1e18 for lock creation as extra protection against dust attacks

### Other changes

- Updated tests to scale amounts to prevent reverting due to new require condition
- Recalculated and updated the expected lock outcomes in tests since we have less precision loss with scaled amounts(used wolframalpha).

### Tested

- Created new tests with 50k locks and PoCs from sherlock submissions to verify new implementation would perform reasonable with large number of locks under various scenerios

### Related issues

- Fixes #407


philbow61 (approved):
LGTM 💯 
how did you come up with the gas base lines in GovernanceIntegration.gas.t.sol?

baroooo:
> LGTM 💯 how did you come up with the gas base lines in GovernanceIntegration.gas.t.sol?

rounded up the gas cost of the same operations with single locks

bayological (approved):

PR #487, opened 2024-08-08
Feat: Make ChainlinkRelayer support implicit pairs

bowd:
### Description

Late last week it became apparent that Chainlink won't provide all derived rates that we need, and that we will need to compute implicit rates on-chain. This PR extends the relayer such that it can aggregate up to 4 chainlink date feeds into a single price. Each data feed can be inverted (1 / p) to make the dominos line up. 

### Other changes

Upgraded the tests to solidity 0.8 and fixed the issue with deploying SortedOracles via the factory by pre-deploying libraries as well because then factory links them automatically when calling `vm.getCode`.

### Tested

Tests have been extended, and I've also used a slightly new pattern that I didn't use before which I like, and might not be evident from reading the code. Specifically for the relayer we have 4 scenarios depending on how many price feeds it aggregates, so I wrote test contracts that extend each other in such a way where test cases from N-1 price feeds get executed also in the test for N price feeds. 

### Related issues

N/A

### Backwards compatibility

Not backwards compatible but not deployed yet.

### Documentation

N/A


contracts/oracles/ChainlinkRelayerV1.sol
```diff
@@ -56,21 +73,75 @@ contract ChainlinkRelayerV1 is IChainlinkRelayer {
...
+   * aggregator followed by one with a set address, resulting in a gap.
+   */
+  error PricePathHasGaps();
 
   /**
    * @notice Initializes the contract and sets immutable parameters.
    * @param _rateFeedId ID of the rate feed this relayer instance relays for.
    * @param _sortedOracles Address of the SortedOracles contract to relay to.
-   * @param _chainlinkAggregator Address of the Chainlink price feed to fetch data from.
+   * @param _chainlinkAggregator0 Addresses of the Chainlink price feeds to fetch data from.
```
chapati23, line 96:
```suggestion
   * @param _maxTimestampSpread Max difference in milliseconds between the earliest and latest timestamp of all aggregators in the price path
   * @param _chainlinkAggregator0 Addresses of the Chainlink price feeds to fetch data from.
```

test/oracles/ChainlinkRelayer.t.sol
```diff
@@ -1,41 +1,193 @@
...
   bytes constant EXPIRED_TIMESTAMP_ERROR = abi.encodeWithSignature("ExpiredTimestamp()");
   bytes constant NEGATIVE_PRICE_ERROR = abi.encodeWithSignature("NegativePrice()");
-  SortedOracles sortedOracles;
-  MockAggregatorV3 chainlinkAggregator;
-  IChainlinkRelayer relayer;
-  address rateFeedId = actor("rateFeed");
-  int256 aPrice = 420000000;
-  uint256 expectedReport = 4200000000000000000000000;
+  bytes constant ZERO_PRICE_ERROR = abi.encodeWithSignature("ZeroPrice()");
+  bytes constant TIMESTAMP_SPREAD_TOO_HUGH = abi.encodeWithSignature("TimestampSpreadTooHigh()");
```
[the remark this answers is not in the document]
chapati23: Also, this isn't used anywhere as far as I can see. Should we add a test for this?

.gitmodules
```diff
@@ -19,3 +19,7 @@
 [submodule "lib/foundry-chainlink-toolkit"]
 	path = lib/foundry-chainlink-toolkit
 	url = https://github.com/smartcontractkit/foundry-chainlink-toolkit
+[submodule "lib/prb-math"]
```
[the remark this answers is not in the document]
bowd: Yes they do similar things, the major issue with Fixidity is that it was abandoned and doesn't support solidity 0.8. Benefits of prb-math are that it's maintained, modern and better tested. It makes use of new solidity language features, like free-floating functions and global operator overloading, to make it gas efficient and to pull into the contract the minimum amount of code needed to interact with the types. Some other general benefits that we're not making use of here is that it gives you access to also signed versions and different sizes (2x18, 21x18, etc) for further optimisations where needed.

contracts/oracles/ChainlinkRelayerV1.sol
```diff
@@ -29,7 +30,7 @@ interface ISortedOraclesMin {
  * @notice The ChainlinkRelayer relays rate feed data from a Chainlink price feed to
  * the SortedOracles contract. A separate instance should be deployed for each
  * rate feed.
- * @dev Assumes that it itself is the only reporter for the given SortedOracles
+ * @dev Assumes that it itself is the only reporter 3or the given SortedOracles
```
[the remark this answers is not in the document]
bowd: 100% 😹

contracts/oracles/ChainlinkRelayerV1.sol
```diff
@@ -83,49 +154,107 @@ contract ChainlinkRelayerV1 is IChainlinkRelayer {
...
+    config.invertAggregator3 = invertAggregator3;
   }
 
   /**
    * @notice Checks if a Chainlink price's timestamp would be expired in SortedOracles.
    * @param timestamp The timestamp returned by the Chainlink aggregator.
    * @return `true` if expired based on SortedOracles expiry parameter.
    */
-  function _isTimestampExpired(uint256 timestamp) internal view returns (bool) {
+  function isTimestampExpired(uint256 timestamp) internal view returns (bool) {
```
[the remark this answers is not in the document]
bowd: hmm it doesn't for me, or at least not when I run `yarn solhint` does it complain in the editor?

contracts/oracles/ChainlinkRelayerV1.sol
```diff
@@ -83,49 +154,107 @@ contract ChainlinkRelayerV1 is IChainlinkRelayer {
...
-      revert ExpiredTimestamp();
+    if (maxTimestamp - minTimestamp > maxTimestampSpread) {
+      revert TimestampSpreadTooHigh();
     }
 
-    if (price < 0) {
-      revert NegativePrice();
+    uint256 lastTimestamp = _sortedOracles.medianTimestamp(rateFeedId);
+
+    if (lastTimestamp > 0 && maxTimestamp <= lastTimestamp) {
```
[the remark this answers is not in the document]
bowd: we will because `maxTimestamp` is the most recent one, so if at 
t0 we report rateA, rateB and rateC
and then rateB has a new report and thus a timestamp > t0
it will be the maxTimestamp and, as long as the spread is ok, it will report.

.gitmodules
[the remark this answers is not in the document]
chapati23: thanks! just watched this which was also a good overview https://www.youtube.com/watch?v=dqbN7WlhCLg

contracts/oracles/ChainlinkRelayerV1.sol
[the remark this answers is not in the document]
chapati23: i also stumbled over this one initially, maybe sth like `oldestTimestamp` and `newestTimestamp` would be slightly more intuitive names here?

contracts/oracles/ChainlinkRelayerV1.sol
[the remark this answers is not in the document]
chapati23: can't reproduce it anymore, think i fiddled locally with stricter solhint settings which I later reverted

contracts/oracles/ChainlinkRelayerV1.sol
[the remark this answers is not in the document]
bowd: yeah, and also lastTimestamp should be something else.

test/oracles/ChainlinkRelayer.t.sol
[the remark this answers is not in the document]
bowd: yup, missed that, added tests now

contracts/oracles/ChainlinkRelayerV1.sol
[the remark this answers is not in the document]
bowd: @philbow61 @chapati23 did a rename pass let me know if it feels better

contracts/oracles/ChainlinkRelayerV1.sol
```diff
@@ -83,49 +156,106 @@ contract ChainlinkRelayerV1 is IChainlinkRelayer {
...
 
-    if (lastTimestamp > 0) {
-      if (timestamp <= lastTimestamp) {
-        revert TimestampNotNew();
+    if (chainlinkAggregator1 != address(0)) {
+      UD60x18 nextReport;
+      (nextReport, timestamp) = readChainlinkAggregator(chainlinkAggregator1, invertAggregator1);
+      report = report.mul(nextReport);
+      oldestChainlinkTs = timestamp < oldestChainlinkTs ? timestamp : oldestChainlinkTs;
+      newestChainlinkTs = timestamp > newestChainlinkTs ? timestamp : newestChainlinkTs;
```
m-chrzan, line 168: If we can't use a static array to clean this up, other approaches could be:
* an `aggregateChainlinkReport` function that passes around oldest/newest timestamp and the aggregated report, so that each branch of this `if-if-if` has just one line of logic
* or, with a slight gas penalty, allocate memory arrays and iterate over that

contracts/oracles/ChainlinkRelayerV1.sol
```diff
@@ -83,49 +156,106 @@ contract ChainlinkRelayerV1 is IChainlinkRelayer {
...
     // This contract is built for a setup where it is the only reporter for the
     // given `rateFeedId`. As such, we don't need to compute and provide
     // `lesserKey`/`greaterKey` each time, the "null pointer" `address(0)` will
     // correctly place the report in SortedOracles' sorted linked list.
-    ISortedOraclesMin(sortedOracles).report(rateFeedId, report, address(0), address(0));
+    ISortedOraclesMin(sortedOracles).report(rateFeedId, reportValue, address(0), address(0));
+  }
+
+  /**
+   * @notice Read and validate a chainlink report from an aggregator.
```
m-chrzan, line 207:
```suggestion
   * @notice Read and validate a Chainlink report from an aggregator.
```

contracts/oracles/ChainlinkRelayerV1.sol
```diff
@@ -83,49 +156,106 @@ contract ChainlinkRelayerV1 is IChainlinkRelayer {
...
+   * It inverts the value if necessary
```
m-chrzan, line 208:
```suggestion
   * It inverts the value if necessary.
```

contracts/oracles/ChainlinkRelayerV1.sol
```diff
@@ -83,49 +156,106 @@ contract ChainlinkRelayerV1 is IChainlinkRelayer {
...
+   * @return price UD60x18 report value
+   * @return timestamp uint256 timestamp of the report
+   */
+  function readChainlinkAggregator(address aggregator, bool invert) internal view returns (UD60x18, uint256) {
+    (, int256 _price, , uint256 timestamp, ) = AggregatorV3Interface(aggregator).latestRoundData();
+    if (_price < 0) {
+      revert NegativePrice();
+    }
+    if (_price == 0) {
+      revert ZeroPrice();
```
m-chrzan, line 218: If we want to disallow 0 price, we could have a single `NonPositivePrice` (or `InvalidPrice`) error.

contracts/oracles/ChainlinkRelayerV1.sol
```diff
@@ -83,49 +156,106 @@ contract ChainlinkRelayerV1 is IChainlinkRelayer {
...
     // `lesserKey`/`greaterKey` each time, the "null pointer" `address(0)` will
     // correctly place the report in SortedOracles' sorted linked list.
-    ISortedOraclesMin(sortedOracles).report(rateFeedId, report, address(0), address(0));
+    ISortedOraclesMin(sortedOracles).report(rateFeedId, reportValue, address(0), address(0));
+  }
+
+  /**
+   * @notice Read and validate a chainlink report from an aggregator.
+   * It inverts the value if necessary
+   * @return price UD60x18 report value
```
m-chrzan, line 209:
```suggestion
   * @return price UD60x18 report value.
```

contracts/oracles/ChainlinkRelayerV1.sol
```diff
@@ -83,49 +156,106 @@ contract ChainlinkRelayerV1 is IChainlinkRelayer {
...
+   * @return timestamp uint256 timestamp of the report
```
m-chrzan, line 210:
```suggestion
   * @return timestamp uint256 timestamp of the report.
```

contracts/oracles/ChainlinkRelayerV1.sol
```diff
@@ -83,49 +156,106 @@ contract ChainlinkRelayerV1 is IChainlinkRelayer {
...
+    }
+    UD60x18 price = chainlinkToUD60x18(_price, aggregator);
+    if (invert) {
+      price = price.inv();
+    }
+    return (price, timestamp);
+  }
+
+  /**
+   * @notice Get the relayer configuration as a Config struct
```
m-chrzan, line 228:
```suggestion
   * @notice Get the relayer configuration as a Config struct.
```

contracts/oracles/ChainlinkRelayerV1.sol
```diff
@@ -83,49 +156,106 @@ contract ChainlinkRelayerV1 is IChainlinkRelayer {
...
+   * @return config Config structure populated with all relayer config fields
```
m-chrzan, line 229:
```suggestion
   * @return config Config structure populated with all relayer config fields.
```

remappings.txt
```diff
@@ -4,4 +4,5 @@ forge-std/=lib/celo-foundry/lib/forge-std/src/
 ds-test/=lib/celo-foundry/lib/forge-std/lib/ds-test/src/
 test/=test/
 contracts/=contracts/
-safe-contracts/=lib/safe-contracts/
\ No newline at end of file
+safe-contracts/=lib/safe-contracts/
+@prb/math/=lib/prb-math/
```
m-chrzan, line 8: Consistency nit: why the `@` prefix if we don't use it for other dependencies? Also, looks like for most we point directly to the contracts directory (so could add `src/`).

test/oracles/ChainlinkRelayer.t.sol
```diff
@@ -1,41 +1,193 @@
...
+  function setUpAggregators() public {
+    chainlinkAggregator0.setRoundData(int256(aggregatorPrice0), uint256(block.timestamp));
+    chainlinkAggregator1.setRoundData(int256(aggregatorPrice1), uint256(block.timestamp));
+    chainlinkAggregator2.setRoundData(int256(aggregatorPrice2), uint256(block.timestamp));
+    chainlinkAggregator3.setRoundData(int256(aggregatorPrice3), uint256(block.timestamp));
+  }
+}
+
+contract ChainlinkRelayerTest_constructor_invalid is ChainlinkRelayerTest {
+  function test_constructorRevertsWhenPathInvalid() public {
```
m-chrzan, line 163: Could be nice to have tests with the other two possible gap locations.

contracts/oracles/ChainlinkRelayerV1.sol
```diff
@@ -44,9 +45,25 @@ contract ChainlinkRelayerV1 is IChainlinkRelayer {
 
   /// @notice The address of the SortedOracles contract to report to.
   address public immutable sortedOracles;
-
-  /// @notice The address of the Chainlink aggregator this contract fetches data from.
-  address public immutable chainlinkAggregator;
+  /**
+   * @notice The addresses and invert settings of the Chainlink aggregators this
+   * contract fetches data from, it's limited to 4 potential path segments,
+   * because we can't have dynamic types as immutable, and it's not worth the gas
```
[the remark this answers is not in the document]
bowd: Immutable args must be only value types (32bytes) so that they're inlined during construction. Not even static length arrays are allowed, although theoretically the compiler could do unrolling and make it work, but it doesn't at the moment. So arrays would have only worked with storage.

contracts/oracles/ChainlinkRelayerV1.sol
bowd: Yeah I kinda went through all those options before setting on this, as it felt the cleanest somehow, but happy to revisit. I felt like the array option wasn't worth it, but happy to add a helper function for the aggregation if you think that's more readable.

remappings.txt
bowd: I think I just went with the install instructions from their repo so I could copy-pasta import statements from their examples 😂, but yeah makes sense to keep it consistent.

contracts/oracles/ChainlinkRelayerV1.sol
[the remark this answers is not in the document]
chapati23: yeah reads better now imo!

chapati23 (approved):

test/oracles/ChainlinkRelayer.t.sol
bowd: There are 12 invalid permutations if I'm not mistaked (2^4 - 4) but I added a few more.

contracts/oracles/ChainlinkRelayerV1.sol
```diff
@@ -83,49 +151,146 @@ contract ChainlinkRelayerV1 is IChainlinkRelayer {
...
+          (report, oldestChainlinkTs, newestChainlinkTs) = combineReports(
+            report,
+            oldestChainlinkTs,
+            newestChainlinkTs,
+            chainlinkAggregator3,
+            invertAggregator3
+          );
+        }
       }
     }
```
bowd, line 184: @m-chrzan do you think this reads better? I'm torn imo. 
@chapati23 / @philbow61 what's your take on this little style bikeshedding 😂

chapati23: feels slightly cleaner to me but tbh given we can't use arrays and need to hardcode it's gonna look a bit ugly either way

contracts/oracles/ChainlinkRelayerV1.sol
m-chrzan: Yeah, I think that would make things much cleaner, plus avoid duplication. Make things easier for our future selves if we ever need to ponder or modify this code.

contracts/oracles/ChainlinkRelayerV1.sol
m-chrzan: I like it. It's technically more literal lines of code, but less logic per branch, no duplication, which IMO is much easier to reason about.

Could even use `combineReports(<ud60x18 1>, uint256.max, 0, aggregator0, invert0)` before the `if` for total uniformity.

test/oracles/ChainlinkRelayerV1.t.sol
```diff
@@ -0,0 +1,605 @@
...
+  }
+
+  function setUp() public override {
+    super.setUp();
+    setUpRelayer();
+    setUpExpectations();
+    setAggregatorPrices();
+  }
+
+  function test_relaysTheRate() public {
```
baroooo, line 342: hmmmm this runs for all setups.. smart :D

baroooo (commented):
looks good in general
I added some comments but some of them are not related to this specific PR

contracts/oracles/ChainlinkRelayerV1.sol
```diff
@@ -26,106 +27,256 @@ interface ISortedOraclesMin {
...
+   * @notice Human-readable description of the rate feed.
+   * @dev Should only be used off-chain for easier debugging / UI generation,
+   * thus the only storage related gas spend occurs in the constructor.
+   */
+  string public rateFeedDescription;
+
+  /// @notice Used when an empty array of aggregators is passed into the constructor.
+  error NoAggregators();
+
+  /// @notice Used when an empty array of aggregators is passed into the constructor.
```
nvtaveras, line 99:
```suggestion
  /// @notice Used when more than four aggregators are passed into the constructor.
```

contracts/oracles/ChainlinkRelayerV1.sol
```diff
@@ -26,106 +27,256 @@ interface ISortedOraclesMin {
...
    */
-  function _isTimestampExpired(uint256 timestamp) internal view returns (bool) {
+  function isTimestampExpired(uint256 timestamp) internal view returns (bool) {
     return block.timestamp - timestamp >= ISortedOraclesMin(sortedOracles).getTokenReportExpirySeconds(rateFeedId);
   }
 
   /**
-   * @notice Converts a Chainlink price to an unwrapped Fixidity value.
+   * @notice Converts a Chainlink price to a UD60x18 value.
    * @param price An price from the Chainlink aggregator.
```
nvtaveras, line 275:
```suggestion
   * @param price A price from the Chainlink aggregator.
```

test/oracles/ChainlinkRelayerFactory.t.sol
```diff
@@ -267,11 +314,11 @@ contract ChainlinkRelayerFactoryTest_getRelayers is ChainlinkRelayerFactoryTest
...
-  function setUp() public {
+  function setUp() public override {
     super.setUp();
 
     vm.prank(owner);
-    relayerAddress = relayerFactory.deployRelayer(aRateFeed, mockAggregator);
+    relayerAddress = relayerFactory.deployRelayer(rateFeeds[0], aRateFeedDescription, 0, oneAggregator(0));
   }
 
   function test_removesTheRelayer() public {
```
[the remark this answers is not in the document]
bowd: I think the line is outdated but why can it be removed?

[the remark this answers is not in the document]
baroooo: the line was pranking for a view call but I can not find it anymore :D

baroooo (approved):

PR #494, opened 2024-08-09
feat: implement a relayer version that uses in-memory arrays

bowd:
Side-hustle of #487 to turn immutable variables into an in-memory array for aggregators.


bowd:
@m-chrzan @philbow61 @chapati23 

So I wanted to test out the ergonomics of this option as well, the gas-penalty compared to not using arrays is around <2k gas which is <1% of gas cost for transactions. We still get the silly nested ifs but a bit more isolated and just in order to build the array. What do y'all think?

contracts/interfaces/IChainlinkRelayer.sol
```diff
@@ -24,6 +24,16 @@ interface IChainlinkRelayer {
     bool invertAggregator3;
   }
 
+  /**
+   * @notice Struct used to represent a segment in the price path.
+   * @member aggregator The address of the Chainink aggregator.
+   * @member invert Wether to invert the aggregator.
```
chapati23, line 30: would find this slightly clearer, especially when I didn't have context 

```suggestion
   * @member invert Wether to invert the aggregator's price feed, i.e. convert CELO/USD to USD/CELO.
```

contracts/oracles/ChainlinkRelayerV2.sol
```diff
@@ -0,0 +1,280 @@
...
+
+  function getTokenReportExpirySeconds(address rateFeedId) external view returns (uint256);
+}
+
+/**
+ * @title ChainlinkRelayer
+ * @notice The ChainlinkRelayer relays rate feed data from a Chainlink price feed to
+ * the SortedOracles contract. A separate instance should be deployed for each
+ * rate feed.
+ * @dev Assumes that it itself is the only reporter for the given SortedOracles
```
chapati23, line 33:
```suggestion
 * @dev Assumes that it is the only reporter for the given SortedOracles
```

contracts/oracles/ChainlinkRelayerV2.sol
```diff
@@ -0,0 +1,280 @@
...
+   */
+  address public immutable chainlinkAggregator0;
+  address public immutable chainlinkAggregator1;
+  address public immutable chainlinkAggregator2;
+  address public immutable chainlinkAggregator3;
+  bool public immutable invertAggregator0;
+  bool public immutable invertAggregator1;
+  bool public immutable invertAggregator2;
+  bool public immutable invertAggregator3;
+  uint256 public immutable aggregatorsCount;
```
chapati23, line 62: nit: think "aggregatorsCount" is like double plural
```suggestion
  uint256 public immutable aggregatorCount;
```

chapati23 (commented):
gg, think this approach reads a bit better to my eyes, and gas penalty seems to be negligible so i'd prefer this one

only smol nit comments

test/oracles/ChainlinkRelayerGasComparison.t.sol
```diff
@@ -0,0 +1,283 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
```
m-chrzan, line 1: Does this need to remain in the repo?

m-chrzan (approved):
Looks good. No strong preference from me on this vs. the version with the helper function, agree that the gas cost increase is small enough. So if you're leaning towards this one, sounds good to me!

contracts/oracles/ChainlinkRelayerV1.sol
```diff
@@ -206,73 +198,42 @@ contract ChainlinkRelayerV1 is IChainlinkRelayer {
...
-    config.chainlinkAggregator1 = chainlinkAggregator1;
-    config.chainlinkAggregator2 = chainlinkAggregator2;
-    config.chainlinkAggregator3 = chainlinkAggregator3;
-    config.invertAggregator0 = invertAggregator0;
-    config.invertAggregator1 = invertAggregator1;
-    config.invertAggregator2 = invertAggregator2;
-    config.invertAggregator3 = invertAggregator3;
+  function getAggregatorsArray() internal view returns (ChainlinkAggregator[] memory aggregators) {
+    aggregators = new ChainlinkAggregator[](aggregatorCount);
+    unchecked {
```
chapati23, line 241: stupid question maybe but why does this need to be `unchecked` ?

always thought this only saved gas for arithmetic operations of which I don't see any here

test/oracles/ChainlinkRelayerFactory.t.sol
```diff
@@ -22,48 +22,49 @@ contract ChainlinkRelayerFactoryTest is BaseTest {
...
     address indexed relayerAddress,
     address indexed rateFeedId,
-    IChainlinkRelayer.Config relayerConfig
+    string rateFeedDescription,
+    IChainlinkRelayer.ChainlinkAggregator[] aggregators
   );
   event RelayerRemoved(address indexed relayerAddress, address indexed rateFeedId);
 
+  function oneAggregator(uint256 aggregatorIndex)
+    internal
```
chapati23, line 36:
```suggestion
    internal view
```

test/oracles/ChainlinkRelayerFactory.t.sol
```diff
@@ -22,48 +22,49 @@ contract ChainlinkRelayerFactoryTest is BaseTest {
...
+    returns (IChainlinkRelayer.ChainlinkAggregator[] memory aggregators)
+  {
+    aggregators = new IChainlinkRelayer.ChainlinkAggregator[](1);
+    aggregators[0] = IChainlinkRelayer.ChainlinkAggregator(mockAggregators[aggregatorIndex], false);
+  }
+
+  function fourAggregators() internal returns (IChainlinkRelayer.ChainlinkAggregator[] memory aggregators) {
```
chapati23, line 43:
```suggestion
  function fourAggregators() internal view returns (IChainlinkRelayer.ChainlinkAggregator[] memory aggregators) {
```

test/oracles/ChainlinkRelayerFactory.t.sol
```diff
@@ -22,48 +22,49 @@ contract ChainlinkRelayerFactoryTest is BaseTest {
...
     relayerFactory = IChainlinkRelayerFactory(new ChainlinkRelayerFactory(false));
     vm.prank(owner);
     relayerFactory.initialize(mockSortedOracles);
   }
 
-  function assertRelayerMatchesConfig(IChainlinkRelayer relayer, IChainlinkRelayer.Config memory expected) internal {
-    IChainlinkRelayer.Config memory actual = relayer.getConfig();
-    assertEq(keccak256(abi.encode(expected)), keccak256(abi.encode(actual)));
-  }
+  // function assertRelayerMatchesConfig(IChainlinkRelayer relayer, IChainlinkRelayer.Config memory expected) internal {
```
chapati23, line 57: can this re removed?

chapati23 (approved):
even cleaner now, just my usual comment nits and some questions for my understanding

contracts/oracles/ChainlinkRelayerV1.sol
bowd: It also applies to array out-of-bound checks and we're essentially already doing those ourselves there. It's a small win, yes, but the biggest victory is my personal satisfaction in getting to use `unchecked`. 😂

chapati23: ah, i see, thanks!

contracts/oracles/ChainlinkRelayerV1.sol
```diff
@@ -27,44 +27,78 @@ interface ISortedOraclesMin {
...
- * rate feed.
- * @dev Assumes that it itself is the only reporter for the given SortedOracles
- * feed.
+ * @notice The ChainlinkRelayer relays rate feed data from a Chainlink price feed, or
+ * an aggregation of multiple Chainlink price feeds to the SortedOracles contract.
+ * A separate instance should be deployed for each rate feed.
+ * @dev Assumes that it is the only reporter for the given SortedOracles feed.
+ * This contract aggregates multiple Chainlink price feeds in order to provide derived rate feeds
+ * to the rest of the protocol. This is needed because it is more efficient for oracle providers
+ * to report FX rates agains the dollar and crypto-asset rates against against the dollar,
```
chapati23, line 36:
```suggestion
 * to report FX rates against the dollar and crypto-asset rates against the dollar,
```

contracts/oracles/ChainlinkRelayerV1.sol
```diff
@@ -27,44 +27,78 @@ interface ISortedOraclesMin {
...
+ * instead of all possible combinations.
+ * For example, for the Philipinean Peso, Chainlink reports PHP/USD, but does not report CELO/PHP
```
chapati23, line 38:
```suggestion
 * For example, for the Philippine Peso, Chainlink reports PHP/USD, but does not report CELO/PHP
```

chapati23 (approved):
2 last typos then good to go!

PR #498, opened 2024-08-19
feat/cosmetic: Move Proxy Contracts around

bowd:
### Description

When we restructured the repo a while ago I moved the proxy contracts to the legacy folder thinking that we might not deploy named proxies in the future, and maybe upgrade to a more modern Proxy implementation, but we have continued to add the proxies inside the `legacy` folder which doesn't feel great as that folder is meant to be removed (and probably can be at this point). 

So in order to embrace what we're doing I propose collocating the proxy contracts to the `contracts/tokens` directory. And as an addition to that, I propose collocating all proxies with their contracts like Martin also did in the `contracts/oracles` for the new ChainlinkRelayerFactory contract and proxy. So I also moved around the proxies in `contracts/proxies`.

This shouldn't really effect anything else.

### Other changes

- Renamed stabletokenPSO to stabletokenPHP

### Tested

Ran tests, all is well.

### Related issues

N/A

### Backwards compatibility

YES

### Documentation

N/A


bayological (approved):
On the fence about having the proxies next to the implementation. I think this would make more sense for the interfaces than proxies since there's never much need to refer back to the proxies, so locating them easily doesn't seem like a problem. On the other hand, we don't have many so not much overcrowding in the folder.

contracts/tokens/StableTokenPSOProxy.sol
```diff
@@ -1,9 +1,9 @@
 // SPDX-License-Identifier: GPL-3.0-or-later
```
baroooo, line 1: contract name is PHP but file name is still PSO

denviljclarke: That's correct. I've updated it

baroooo (approved):

PR #499, opened 2024-08-21
feat: Add relayerDeployer ACL role to the factory

bowd:
### Description

We realised that it would be better operationally to separate the proxy ownership ACL from the deployer ACL, that way we can keep using the `mento-deployer` account for relayer deploy, and use the multisig only for proxy updates.

### Other changes

N/A

### Tested

Tests added

### Related issues

N/A

### Backwards compatibility

No, but no mainnet deployment yet.

### Documentation

N/A

### DIFF
--- a/contracts/oracles/ChainlinkRelayerFactory.sol
+++ b/contracts/oracles/ChainlinkRelayerFactory.sol
@@ -61,7 +61,7 @@ contract ChainlinkRelayerFactory is IChainlinkRelayerFactory, OwnableUpgradeable
 
   /// @notice Modifier to restrict a function to the deployer.
   modifier onlyDeployer() {
-    if (msg.sender != relayerDeployer) {
+    if (msg.sender != relayerDeployer && msg.sender != owner()) {
       revert NotAllowed();
     }
     _;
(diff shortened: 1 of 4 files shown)


contracts/oracles/ChainlinkRelayerFactory.sol
```diff
@@ -69,9 +85,19 @@ contract ChainlinkRelayerFactory is IChainlinkRelayerFactory, OwnableUpgradeable
...
     sortedOracles = _sortedOracles;
+    relayerDeployer = _relayerDeployer;
+  }
+
+  /**
+   * @notice Sets the address of the relayer deployer.
+   * @param _relayerDeployer The address of the relayer deployer.
+   */
+  function setRelayerDeployer(address _relayerDeployer) external onlyOwner {
+    emit RelayerDeployerUpdated(_relayerDeployer, relayerDeployer);
```
philbow61, line 99: nit: maybe switch emit with state change. I think it looks better but needs extra line 🤔

chapati23 (approved):
looks all correct to me

philbow61 (approved):
LGTM 👍

contracts/oracles/ChainlinkRelayerFactory.sol
```diff
@@ -51,6 +56,17 @@ contract ChainlinkRelayerFactory is IChainlinkRelayerFactory, OwnableUpgradeable
    */
   error NoRelayerForRateFeedId(address rateFeedId);
 
+  /// @notice Thrown when a non-deployer tries to call a deployer-only function.
+  error NotAllowed();
+
+  /// @notice Modifier to restrict a function to the deployer.
+  modifier onlyDeployer() {
+    if (msg.sender != relayerDeployer) {
```
m-chrzan, line 64: Would it make sense to have `owner` also have deployment privileges as a failsafe? Of course it's possible for the owner to just set itself as the deployer and go from there, but could save some headache in an emergency situation if more complex relayer management is necessary to fix something.

test/oracles/ChainlinkRelayerFactory.t.sol
```diff
@@ -24,13 +26,16 @@ contract ChainlinkRelayerFactoryTest is BaseTest {
   address aRateFeed = rateFeeds[0];
   string aRateFeedDescription = "CELO/USD";
 
+  bytes constant NOT_ALLOWED = abi.encodeWithSignature("NotAllowed()");
```
m-chrzan, line 29: Consistency nit: `NOT_ALLOWED_ERROR`, as in ChainlinkRelayer.t.sol.

m-chrzan (approved):
Looks good, just two small suggestions.

PR #501, opened 2024-08-22
THE BIG ONE 🐘 💣 🚝

bowd:
### Description

As I told some of you at dinner, I've had a bad case of insomnia over the last week. This PR resulted from a couple of late night coding sessions and the incessant need to make things nicer in our repos. 

It's a big PR but 𝚫code is negative so 🤷 .

What happens in this mondo-PR: 

1. All `celo` contracts are removed form the repo and replaced with the `@celo/contracts` NPM package. With a small caveat.
2. All `interfaces` are made to work with `0.8` and cover more of the contract functions.
3. All `tests` contracts are updated to `0.8` and use `deployCode` helpers to deploy lower-version contracts, and the use the interfaces to interact with them.
4. We make use of this https://github.com/foundry-rs/foundry/pull/8668 to improve compile time.
5. Update `solhint` and `prettier` to a recent version and fix/ignore all issues.
6. Fix solc compiler warnings.
7. Fix/ignore slither > informational warnings.
8. Slight Refactor of ForkTests to make them better to work with.
9. Restructuring of the tests folder.

##### `@celo/contracts`

The only caveat of using the `@celo/contracts` package is that `UsingRegistry.sol` in there needs to import things from `mento-core` and I didn't manage to get it working ok with remappings, so I kept a copy of `UsingRegistry.sol` in `contracts/common`. It's only used in `swap/Reserve.sol` and when we remove it from there we can completely kill `common`.

##### The foundry WIP PR

The reason it takes so long to compile our code is because we need `via-ir` for `Airgrab.sol` and `StableTokenV2.sol`, and `via-ir` is super slow. But with the compiler restrictions implemented in https://github.com/foundry-rs/foundry/pull/8668 we can have multiple compiler profile settings for subgraphs of the source-graph, which compile in parallel with different settings.

You can easily install that version of foundry locally, (you have to have rust installed tho):
```
foundryup -p 8668
```

With this version of foundry and the settings in `foundry.toml`, if you're not working in a part of the source graph that contains `Airgrab.sol` or `StableTokenV2.sol`, compilation will happen without `via-ir` and will be super snappy. However if you do touch that source graph it will still take noticeably long. Right now on my machine full clean compilation takes 80seconds. It used to take >3minutes. 

##### ForkTest Refactoring

Our fork tests can get a bit heavy because they're running test assertions for all the exchanges in a single test call. I've refactor it a bit and split out exchange assertions into their own tests contracts that need to be manually created when new exchanges are deployed. There's a chain-level assertion on the number of exchanges that helps us catch this and keep them up to date.

### Other changes

> _Describe any minor or "drive-by" changes here._

no minor changes here, no :))

### Tested

Tested? Tested!

### Related issues

Fixes the issues in my head.

### Backwards compatibility

What even is that?

### Documentation

Holy Bible


RyRy79261:
Command `foundryup -p 8668` results in 
image

bowd:
> Command `foundryup -p 8668` results in image

You might have an older version of `foundryup` they made some breaking changes at some point, try running: `curl -L https://foundry.paradigm.xyz | bash` first.

RyRy79261:
`yarn fork-test`

image

RyRy79261:
`yarn fork-test:alfajores` 
image

RyRy79261 (approved):

bowd:
@RyRy79261 tackled the fork tests here: #503 because it felt like a bigger beast and worth keeping separate. The reasons these were failing here weren't related to the changes in the branch, they were just a bit flakey, Hopefully you should see all green on the other PR.

contracts/oracles/ChainlinkRelayerV1.sol
```diff
@@ -287,7 +278,9 @@ contract ChainlinkRelayerV1 is IChainlinkRelayer {
 
     if (otherRate < rate) {
       lesserKey = otherOracle;
+      greaterKey = address(0);
```
baroooo, line 281: NIT: is this here to make linter/slither happy?

baroooo (approved):
🫨 🫨 🚢 🚢 
I noticed it sometime takes time between running the `forge test` and compilation starts(2-3 mins) and also between compilation ends and tests run(2-3 mins). It basically stays there without any feedback. I guess that is caused by the foundry version and won't be an issue after the latest PR of eliminating the need for viaIR.

contracts/oracles/ChainlinkRelayerV1.sol
bowd: yeah it was,  but I just reverted it and added ignores so we don't modify the contract that I just released in v2.5.0. So close.

bowd:
@baroooo yeah I noticed that as well (and commented on the foundry WIP branch about it) but yeah like you said it's not a problem cose it will go away completely with the other PR.

PR #502, opened 2024-08-23
Remove the need for `--via-ir` in our contracts

bowd:
### Description

After getting more and more annoyed with the CI time and other small issues around this I decided to really understand how big is our need for `--via-ir` compilation. Turns out it's really small. I mean really small.

There are just a couple of places where this is causing issues: 

1. `Airgrab#claim` has too many variables. I fixed this with a `struct` for the fractal proof, and an internal function to do the lock. Changing it is ok in my opinion. If we want to reuse this contract we'll have to redeploy it anyway, we have a tag for the version when it was deployed, all is well.

2. `GovernanceFactory#createGovernance` has too many variables. I just broke it down in internal functions. Again, I don't see a problem in changing this, even if it's deployed and etc.

3. `StableTokenV2#initialize` has too many arguments. When we made StableTokenV2 we wanted  to keep the same interface for the `initialize` method to keep the interfaces between stable tokens consistent, but it was actually a bad call. There's no good reason why we have it that way and it's a method that can anyway only be called once. I would say it's ok to also change this without making a StableTokenV3 for it. Basically we will have a few contracts which will have a different signature for the initialize method that anyway can't be called anymore. But this is the one change the feels least nice.

4. A couple of tests with many variables where I used structs for grouping or broke down into internal functions.

What happens afterwards, well:
```
╰─ forge test
[⠊] Compiling...
[⠒] Compiling 64 files with Solc 0.5.17
[⠢] Compiling 55 files with Solc 0.8.26
[⠒] Compiling 223 files with Solc 0.8.18
[⠒] Solc 0.5.17 finished in 307.90ms
[⠘] Solc 0.8.26 finished in 1.75s
[⠒] Solc 0.8.18 finished in 5.50s
Compiler run successful with warnings:
```
The whole fucking shabang runs in 5.5s. Jesus.
CI takes ~1minute instead of ~15minutes.

P.S. Thank you Claude for the bash script.

### Other changes

It changed my life.

### Tested

Thoroughly.

### Related issues

Sleep deprivation.

### Backwards compatibility

Who are you calling backward?

### Documentation

I read it, yes.


contracts/governance/GovernanceFactory.sol
```diff
@@ -248,34 +300,47 @@ contract GovernanceFactory is Ownable {
...
     // =============================================
     // =========== Configure Ownership =============
     // =============================================
     emission.transferOwnership(address(governanceTimelock));
     locking.transferOwnership(address(governanceTimelock));
     proxyAdmin.transferOwnership(address(governanceTimelock));
     mentoToken.transferOwnership(address(governanceTimelock));
+  }
 
-    emit GovernanceCreated(
```
baroooo, line 270: looks like we lost the event, which is not a big deal since we have the addresses saved anyways but I would either remove the event description on L:36 which is not used anymore or emit the event as the last thing in `createGovernance`

contracts/oracles/ChainlinkRelayerV1.sol
```diff
@@ -280,7 +280,9 @@ contract ChainlinkRelayerV1 is IChainlinkRelayer {
 
     if (otherRate < rate) {
       lesserKey = otherOracle;
+      greaterKey = address(0);
```
baroooo, line 283: looks like a rebasing error?

bin/check-contracts.sh
```diff
@@ -0,0 +1,66 @@
+#!/bin/bash
```
baroooo, line 1: I would add this to package.json as well sth like `"check-no-ir": "./bin/check-contracts.sh"`

.github/workflows/lint_test.yaml
```diff
@@ -47,10 +47,8 @@ jobs:
       - name: "Run the tests"
         run: "forge test"
 
-      - name: "Build the contracts"
-        run: |
-          forge --version
-          forge build --sizes --skip test/**/*
+      - name: "Check contract sizes"
+        run: env FOUNDRY_PROFILE=optimized forge build --sizes --skip test/**/*
```
baroooo, line 51: do we want to add this to package json so we run it locally when needed?

baroooo (approved):
✅

chapati23:
image

nice new compile time

chapati23 (approved):
lgtm, just one outdated-looking outcommented line

PR #503, opened 2024-08-24
Improve 🍴 Fork Tests

bowd:
### Description

Another branch PR on THE BIG ONE. This is a long-overdue cleanup and improvement of our fork tests structure. These tests are important and we will want to maintain and grow them, but the structure they were in made this very hard. They were also terribly brittle and fleaky.

The [new structure](https://github.com/mento-protocol/mento-core/blob/81bd18666ea088d6df1cfa461d5f902e36a2c49a/test/fork/README.md) I'm proposing aims to tackle two main issues:
- Fork tests were executing assertions for all exchanges in a single test making it hard to debug and follow.
- Helpers and utility functions weren't structured and were hard to debug and maintain.

Note for reviewers:
99% of the code inside of the functions is just copied from the old structure so I wouldn't do a full review of everything, I would focus on the structure itself. I did make some small tweaks to improve flakiness, for example skipping some of the tests when the limits prevent them, which I think is a fair thing to do, as long as the tests run periodically.

### Other changes

I added a CI job for running fork-tests daily on `develop`. 

### Tested

Yeah, they are, after all, tests.

### Related issues

N/A

### Backwards compatibility

N/A

### Documentation

N/A


test/fork/BaseForkTest.sol
```diff
@@ -34,17 +29,9 @@ interface IMint {
  * However, it should be exausitve in testing invariants across all tradable pairs
  * in the system, therfore each test should.
```
philbow61, line 30:
```suggestion
 * in the system, therefore each test should.
```

baroooo (approved):

Issue #495, opened 2024-08-13
[SPIKE] Solve Factory.redeployRelayer() problem with missing lesserKeys and greaterKeys


chapati23:
### How to reproduce

1. Deploy a relayer on Alfajores
2. Whitelist it in SortedOracles on Alfajores
3. Make at least one report by calling `Relayer.relay()` manually
4. Remove the relayer from SortedOracles again
5. Redeploy the relayer via `Factory.redeployRelayer()`
6. Whitelist the new relayer in SortedOracles
7. Try to call `Relayer.relay()` and see it fail
8. There's a report from the previous relayer, which will never expire. And the new relayer can't report anymore, due to lesser/greater keys being 0.

PR #507, opened 2024-08-28
fix: gas tests

baroooo:
### Description

Context: We create 50k locks to check how much gas is used by governance and locking operations for users with high number of locks. It was working fine until a couple of weeks ago. Then we disabled the gas tests to make CI happy.

The problem was related to a [recent change on foundry](https://github.com/foundry-rs/foundry/pull/8274/files). They reduced the default gas limit for tests from `9223372036854775807` to `1073741824` which is not enough to create as many locks as we want.

This PR fixes the issue by setting the gas limit to the previous default value of `9223372036854775807` or `i64::MAX / 2` and re-enables the gas tests. I don't think this will cause any problems as the only downside of this is if we have an infinite loop in our tests, it will take some time to consume all the available gas before the test fails.

### Other changes

- Used a common var `viewCallGas` instead of creating one for each test
- Increased the expected view call gas usages from 20k to 30k as they consume slightly higher gas now, probably thats because we removed the gas optimization across repo recently

### Tested

✅ 

### Related issues

- Fixes [#495](https://github.com/mento-protocol/mento-general/issues/495)

### DIFF
--- a/foundry.toml
+++ b/foundry.toml
@@ -10,7 +10,9 @@ fuzz_runs = 256
 gas_reports = ["*"]
 optimizer = false
 legacy = true
-no_match_contract = "(ForkTest)|(GovernanceGasTest)"
+no_match_contract = "ForkTest"
+gas_limit = 9223372036854775807
+
 
 allow_paths = [
   "node_modules/@celo"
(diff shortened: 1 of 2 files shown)


RyRy79261 (approved):

chapati23 (approved):
nice work, confirmed it works locally and on CI

PR #523, opened 2024-10-09
feat: add Broker liquidity check

philbow61:
### Description

This Pr adds a check on Broker quotes ensuring that if the TokenOut is a collateral asset the reserve can match the amountOut if not the call reverts. 

### Tested

fixed the test + added additional unit tests 

### Related issues

- Fixes https://github.com/mento-protocol/mento-general/issues/554

### DIFF
--- a/contracts/swap/Broker.sol
+++ b/contracts/swap/Broker.sol
@@ -140,6 +140,10 @@ contract Broker is IBroker, IBrokerAdmin, Initializable, Ownable, ReentrancyGuar
     uint256 amountOut
   ) external view returns (uint256 amountIn) {
     require(isExchangeProvider[exchangeProvider], "ExchangeProvider does not exist");
+    address reserve = exchangeReserve[exchangeProvider];
+    if (IReserve(reserve).isCollateralAsset(tokenOut)) {
+      require(IERC20(tokenOut).balanceOf(reserve) >= amountOut, "Insufficient balance in reserve");
+    }
     amountIn = IExchangeProvider(exchangeProvider).getAmountIn(exchangeId, tokenIn, tokenOut, amountOut);
   }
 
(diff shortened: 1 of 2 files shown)


contracts/swap/Broker.sol
```diff
@@ -161,6 +165,10 @@ contract Broker is IBroker, IBrokerAdmin, Initializable, Ownable, ReentrancyGuar
   ) external view returns (uint256 amountOut) {
     require(isExchangeProvider[exchangeProvider], "ExchangeProvider does not exist");
     amountOut = IExchangeProvider(exchangeProvider).getAmountOut(exchangeId, tokenIn, tokenOut, amountIn);
+    address reserve = exchangeReserve[exchangeProvider];
+    if (IReserve(reserve).isCollateralAsset(tokenOut)) {
```
bayological, line 169: Would be more efficient to extract this check and move it before `exchangeProvider.getAmountOut` ?

contracts/swap/Broker.sol
```diff
@@ -161,6 +165,10 @@ contract Broker is IBroker, IBrokerAdmin, Initializable, Ownable, ReentrancyGuar
...
+      require(IERC20(tokenOut).balanceOf(reserve) >= amountOut, "Insufficient balance in reserve");
```
bayological, line 170: This check could be bypassed by sending enough `tokenOut` to the reserve. Could this be problematic down the line?

bayological (changes requested):

contracts/swap/Broker.sol
chapati23: we need `amountOut` in the require condition, so I think in this case the check can't happen before

wrt extracting I'm indifferent, would be slightly cleaner to DRY this up by creating a little `ensureReserveHasAtLeast(uint256 amount, address token)` helper but would also slightly increase gas cost

bayological: Specifically the isCollateralAsset(tokenOut) check but tbh the gains would be minimal.

test/unit/swap/Broker.t.sol
```diff
@@ -228,7 +228,53 @@ contract BrokerTest_getAmounts is BrokerTest {
     broker.getAmountIn(randomExchangeProvider, exchangeId, address(stableAsset), address(collateralAsset), 1e24);
   }
 
-  function test_getAmountIn_whenExchangeProviderIsSet_shouldReceiveCall() public view {
+  function test_getAmountIn_whenReserveBalanceIsLessThanAmountOut_shouldRevert() public {
```
chapati23, line 231: Even though it's functionally redundant, it could make the code a tad nicer to read if you explicitly minted 0 just to show that the reserve doesn't have enough:

```sol
collateralAsset.mint(address(reserve), 0);
```

chapati23 (approved):
nice work, just some stylistic comments/suggestions for the tests

bayological (approved):

baroooo (approved):

PR #524, opened 2024-10-11
Feat/change gdollar modifiers

chapati23:
### Description
Fixes https://github.com/mento-protocol/mento-general/issues/562
Fixes https://github.com/mento-protocol/mento-general/issues/560

- [x] Changed access controls for GoodDollarExchangeProvider according to https://github.com/mento-protocol/mento-general/issues/562
- [x] Changed access controls for GoodDollarExpansionController according to https://github.com/mento-protocol/mento-general/issues/560
- [x] Updated and expanded unit tests
- [x] Fixed prettier config warnings
- [x] Fixed all linter warnings
- [x] Fixed `Lint & Test` CI job (was missing the rpc URL env var)

### DIFF
--- a/.prettierrc.yml
+++ b/.prettierrc.yml
@@ -20,6 +20,8 @@ overrides:
   # File-specific Config
   - files:
       [
+        contracts/common/IERC20MintableBurnable.sol,
+        contracts/common/SafeERC20MintableBurnable.sol,
         contracts/goodDollar/**/*.sol,
         contracts/governance/**/*.sol,
         contracts/interfaces/*.sol,
@@ -37,8 +39,3 @@ overrides:
       - contracts/interfaces/IExchange.sol
     options:
       compiler: 0.5.17
-  - files:
-      - contracts/common/IERC20MintableBurnable.sol
-      - contracts/common/SafeERC20MintableBurnable.sol
-    options:
-      compiler: 0.8.0


bayological (commented):
I understand these aren’t your changes, but seeing this again raises some concerns.

Why are we introducing GoodDollar-specific code into our codebase? We made efforts in the past to decouple from Celo, and now it seems like we're reintroducing a dependency with another project. 

Unless there's a strong reason against this, the GoodDollar specific code should be a separate library or project that makes use of our existing codebase where necessary instead of being directly integrated. This will help to keep mento core clean, maintainable and easier to evolve independently. 

I'd recommend considering how we can extract the GoodDollar code into a separt repo/project that interfaces with our code, in the same way we now integrate with Celo core.

WDYT? Is there a specific reason for adding everything in this repo? @philbow61 @chapati23

bayological (approved):
Nice work 👍

.prettierrc.yml
```diff
@@ -21,18 +37,8 @@ overrides:
...
-  - files: [test/**/*.sol]
-    options:
-      compiler: ""
-  - files: [contracts/oracles/Chainlink*.sol]
+  - files:
+      - contracts/common/IERC20MintableBurnable.sol
+      - contracts/common/SafeERC20MintableBurnable.sol
     options:
-      compiler: 0.8.18
+      compiler: 0.8.0
```
baroooo, line 44: not sure how prettier uses the compiler but 0.8.18 should be fine for these two files since it is also used to compile them

baroooo (approved):

.prettierrc.yml
chapati23: you're right, think i just copy/pasted from the pragma, but works fine with 0.8.18 too

chapati23:
> I understand these aren’t your changes, but seeing this again raises some concerns.
> 
> Why are we introducing GoodDollar-specific code into our codebase? We made efforts in the past to decouple from Celo, and now it seems like we're reintroducing a dependency with another project.
> 
> Unless there's a strong reason against this, the GoodDollar specific code should be a separate library or project that makes use of our existing codebase where necessary instead of being directly integrated. This will help to keep mento core clean, maintainable and easier to evolve independently.
> 
> I'd recommend considering how we can extract the GoodDollar code into a separt repo/project that interfaces with our code, in the same way we now integrate with Celo core.
> 
> WDYT? Is there a specific reason for adding everything in this repo? @philbow61 @chapati23

it's a fair point. tbh I didn't question this so far assuming a conscious decision had been made to keep it in core.

i think at this point it's a bit late to extract it. i think a better order of things would be:

1. Finish implementation & testing work as is so we can start the audit
2. Decide if/how to extract into another repo

It's not crystal clear IMO because we're also changing making some broker changes and introducing the option to have multiple reserves so some changes _had_to be done in core. But it's a fair question whether the Good$-specific contracts and interfaces need to live in core as well.

Issue #525, opened 2024-10-11
Resolve Slither Warnings on GoodDollar & Bancor Contracts


chapati23:
There are a number of slither warnings on the GoodDollar & Bancor contracts that need to be analyzed and resolved.

### What
- [ ] Run slither (`yarn slither`)
- [ ] Resolve all issues (either by code change for legitimate warnings or adding inline ignore comments to silence them, along with an explanation on why we're silencing them)
- [ ] Make sure CI is green (incl. slither job)

PR #526, opened 2024-10-11
Address Slither Issues in GoodDollar/Bancor Contracts

chapati23:
### Description

Fixes https://github.com/mento-protocol/mento-core/issues/525

Mostly false positives where I added comments.

One change I did make was adding reentrancy guards to the `mint` functions in the `GoodDollarExpansionController`. 

Even though the slither warnings were only a "medium" around possibly manipulating event data via re-entrancy, I thought it'd still be a good idea to disallow reentrancy for all mint functions.

### DIFF
--- a/.github/workflows/lint_test.yaml
+++ b/.github/workflows/lint_test.yaml
@@ -1,55 +1,59 @@
-name: "CI"
+name: CI
 
 env:
-  FOUNDRY_PROFILE: "ci"
+  FOUNDRY_PROFILE: ci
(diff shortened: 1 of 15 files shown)


baroooo (commented):
looks good, but I don't think ReentrancyGuard provides any value here since the functions: 
1- do not give the right of execution to any arbitrary contract, calls are made to known erc20 tokens and mento protocol contracts
2- follows check-effects-interactions pattern correctly

PR #528, opened 2024-10-15
Streamline comments and errors

chapati23:
### Description

Fixes https://github.com/mento-protocol/mento-general/issues/552

- [x] Made error messages across GoodDollar contracts more consistent
- [x] Made comments across GoodDollar comments more consistent
- [x] Introduced @inheritdoc to reduce comment duplication between interfaces and implementations in 0.8 contracts
- [x] Changed `Broker.configureTradingLimit()` from ~~public~~ to **external**

### DIFF
--- a/contracts/goodDollar/interfaces/IGoodProtocol.sol
+++ b/contracts/goodDollar/interfaces/IGoodProtocol.sol
@@ -9,7 +9,6 @@ interface IGoodDollar {
 
   function safeTransferFrom(address from, address to, uint256 value) external;
 
-  // Only used in Fork Tests to give the Broker minting rights during testing
   function addMinter(address _minter) external;
 
   function isMinter(address account) external view returns (bool);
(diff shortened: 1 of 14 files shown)


contracts/goodDollar/interfaces/IGoodProtocol.sol
```diff
@@ -9,6 +9,7 @@ interface IGoodDollar {
 
   function safeTransferFrom(address from, address to, uint256 value) external;
 
+  // Only used in Fork Tests to give the Broker minting rights during testing
```
baroooo, line 12: not sure if this comment makes sense since it is specific to how we use the interface

baroooo (approved):
LGTM ✅

bowd (approved):
❤️

contracts/goodDollar/interfaces/IGoodProtocol.sol
chapati23, line 12:
```suggestion
```

chapati23: yeah, agreed, was just a stream-of-thought comment for me I think, removed

PR #529, opened 2024-10-16
Feat(454): Exchange provider tests 

RyRy79261:
### Description

This PR adds additional tests to to the exchange provider and includes fuzz testing

### Related issues

- Fixes https://github.com/mento-protocol/mento-general/issues/454
-

### DIFF
--- a/test/unit/goodDollar/GoodDollarExchangeProvider.t.sol
+++ b/test/unit/goodDollar/GoodDollarExchangeProvider.t.sol
@@ -10,6 +10,7 @@ import { ERC20 } from "openzeppelin-contracts-next/contracts/token/ERC20/ERC20.s
 import { IReserve } from "contracts/interfaces/IReserve.sol";
 import { IExchangeProvider } from "contracts/interfaces/IExchangeProvider.sol";
 import { IBancorExchangeProvider } from "contracts/interfaces/IBancorExchangeProvider.sol";
+import { UD60x18, unwrap, wrap } from "prb/math/UD60x18.sol";
 
 contract GoodDollarExchangeProviderTest is Test {
   /* ------- Events from IGoodDollarExchangeProvider ------- */
@@ -300,6 +301,24 @@ contract GoodDollarExchangeProviderTest_mintFromExpansion is GoodDollarExchangeP
     exchangeId = exchangeProvider.createExchange(poolExchange1);
   }
 
+  function calculateExpectedMintFromExpansion(
+    uint256 tokenSupply,
+    uint32 reserveRatio,
+    uint256 expansionScaler
+  ) public pure returns (uint256 expectedAmountToMint) {
+    require(expansionScaler > 0, "Expansion rate must be greater than 0");
+
+    // Convert to UD60x18 for precise calculations
+    UD60x18 scaledRatio = wrap(uint256(reserveRatio) * 1e10);
+    UD60x18 newRatio = scaledRatio.mul(wrap(expansionScaler));
+
+    UD60x18 numerator = wrap(tokenSupply).mul(scaledRatio);
+    numerator = numerator.sub(wrap(tokenSupply).mul(newRatio));
+
+    expectedAmountToMint = unwrap(numerator.div(newRatio));
+    return expectedAmountToMint;
+  }
+
   function test_mintFromExpansion_whenCallerIsNotExpansionController_shouldRevert() public {
     vm.prank(makeAddr("NotExpansionController"));
     vm.expectRevert("Only ExpansionController can call this function");
@@ -343,16 +362,19 @@ contract GoodDollarExchangeProviderTest_mintFromExpansion is GoodDollarExchangeP
 
   function testFuzz_mintFromExpansion(uint256 expansionRate) public {
     // Assume a valid expansion rate between 1 and 99.99%
-    vm.assume(expansionRate > 0 && expansionRate < 1e18);
+    vm.assume(expansionRate > 3 && expansionRate < 1e18);
 
     uint256 initialTokenSupply = poolExchange1.tokenSupply;
     uint32 initialReserveRatio = poolExchange1.reserveRatio;
 
-    // Calculate expected values
-    uint256 newRatio = (uint256(initialReserveRatio) * expansionRate) / 1e18;
-    uint256 expectedAmountToMint = (initialTokenSupply * uint256(initialReserveRatio) - initialTokenSupply * newRatio) /
-      newRatio;
+    // Calculate expected values using the helper function
+    uint256 expectedAmountToMint = calculateExpectedMintFromExpansion(
+      initialTokenSupply,
+      initialReserveRatio,
+      expansionRate
+    );
 
+    uint256 newRatio = (uint256(initialReserveRatio) * expansionRate) / 1e18;
     uint32 expectedReserveRatio = uint32(newRatio);
 
     vm.expectEmit(true, true, true, true);
--- a/test/unit/goodDollar/GoodDollarExchangeProvider.t.sol
+++ b/test/unit/goodDollar/GoodDollarExchangeProvider.t.sol
@@ -3,7 +3,7 @@ pragma solidity 0.8.18;
 // solhint-disable func-name-mixedcase, var-name-mixedcase, state-visibility
 // solhint-disable const-name-snakecase, max-states-count, contract-name-camelcase
 
-import { Test, console } from "forge-std/Test.sol";
+import { Test } from "forge-std/Test.sol";
 
 import { GoodDollarExchangeProvider } from "contracts/goodDollar/GoodDollarExchangeProvider.sol";
 import { ERC20 } from "openzeppelin-contracts-next/contracts/token/ERC20/ERC20.sol";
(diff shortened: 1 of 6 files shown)


test/unit/goodDollar/GoodDollarExchangeProvider.t.sol
```diff
@@ -341,6 +361,50 @@ contract GoodDollarExchangeProviderTest_mintFromExpansion is GoodDollarExchangeP
     assertEq(poolExchangeAfter.reserveRatio, expectedReserveRatio);
   }
 
+  function testFuzz_mintFromExpansion(uint256 expansionScaler) public {
+    // Assume a valid expansion scalar >0% and <2%
+    expansionScaler = bound(uint256(expansionScaler), 100, 1e18);
+    // vm.assume(expansionScaler > 3 && expansionScaler < 1e18);
```
bowd, line 367: Delete line

test/unit/goodDollar/GoodDollarExchangeProvider.t.sol
```diff
@@ -341,6 +361,50 @@ contract GoodDollarExchangeProviderTest_mintFromExpansion is GoodDollarExchangeP
     assertEq(poolExchangeAfter.reserveRatio, expectedReserveRatio);
   }
 
+  function testFuzz_mintFromExpansion(uint256 expansionScaler) public {
```
bowd, line 364: This is I guess a conceptual nit pick I have with this test. I dislike the idea of reimplementing logic in test and validating against it. What we're kinda saying here is that we have this logic that calculates the expected mint, and also this logic that determines the new ratio, and it's somewhere in a contract and we're reimplementing it in the test and validating that we've done that correctly. And then we're also doing it in a fuzz test.

TBH I'd do something a bit more easy and general here. Given that we're fuzzing for the expansion scalar, I would assert that, if expansion scalar <100% (i.e. we're making the reserve ration smaller, i.e. we're going to mint more stables) then amountToMint > 0, and I think we can also estimate the amountToMint as a delta around currentSupply * (1/expansionScaler - 1) (off the dome so this might be wrong). And I would also only assert that the newRatio < oldRatio.

test/unit/goodDollar/GoodDollarExchangeProvider.t.sol
bowd, line 367: Delete line

test/unit/goodDollar/GoodDollarExchangeProvider.t.sol
```diff
@@ -341,6 +361,50 @@ contract GoodDollarExchangeProviderTest_mintFromExpansion is GoodDollarExchangeP
...
+    IBancorExchangeProvider.PoolExchange memory poolExchangeAfter = exchangeProvider.getPoolExchange(exchangeId);
+    assertEq(
+      poolExchangeAfter.tokenSupply,
+      initialTokenSupply + amountToMint,
+      "Token supply should increase by minted amount"
+    );
+
+    assertEq(poolExchangeAfter.reserveRatio, expectedReserveRatio, "Reserve ratio should be updated correctly");
+  }
+  function testMintFromExpansion_RevertWhenNewRatioIsZero() public {
```
bowd, line 399: Missing a new line

test/unit/goodDollar/GoodDollarExchangeProvider.t.sol
```diff
@@ -351,6 +415,45 @@ contract GoodDollarExchangeProviderTest_mintFromExpansion is GoodDollarExchangeP
...
+    uint256 totalMinted = 0;
+    for (uint256 i = 0; i < 5; i++) {
+      uint256 amountToMint = exchangeProvider.mintFromExpansion(exchangeId, 0.99e18);
+      totalMinted += amountToMint;
+    }
+    vm.stopPrank();
+
+    IBancorExchangeProvider.PoolExchange memory poolExchangeAfter = exchangeProvider.getPoolExchange(exchangeId);
+    assertEq(poolExchangeAfter.tokenSupply, poolExchange1.tokenSupply + totalMinted);
+    // Add more assertions to check the final state
```
bowd, line 436: Should we?

test/unit/goodDollar/GoodDollarExchangeProvider.t.sol
```diff
@@ -351,6 +415,45 @@ contract GoodDollarExchangeProviderTest_mintFromExpansion is GoodDollarExchangeP
...
+
+  function test_mintFromExpansion_withMaximumExpansion() public {
+    uint256 maxExpansionScaler = 1e18 - 1; // Just below 100%
+
+    vm.prank(expansionControllerAddress);
+    uint256 amountToMint = exchangeProvider.mintFromExpansion(exchangeId, maxExpansionScaler);
+
+    assertGt(amountToMint, 0);
+    // Add more assertions to check the resulting state
+  }
```
bowd, line 456: Just bellow 100% is minimum expansion. maximum expansion is the smallest possible expansionScalar. Outside of this PR but I don't really like the naming here. 
I think `expansionScaler` should be `reserveRatioScaler` or `reserveRatioScaleFactor`.

bowd (changes requested):

test/unit/goodDollar/GoodDollarExchangeProvider.t.sol
RyRy79261: I did mention this in the discussions, but because we haven't got any external contracts which rely on the output value to be accurate, or lining up with expectations externally. 

Is your recommendation to: 

```
if expansion < 100% && amountToMint > 0 {
   assertApproximateEq: amountToMint, (currentSupply * (1/expansionScalar - 1));
}
```

test/unit/goodDollar/GoodDollarExchangeProvider.t.sol
```diff
@@ -408,6 +511,70 @@ contract GoodDollarExchangeProviderTest_mintFromInterest is GoodDollarExchangePr
...
+    IBancorExchangeProvider.PoolExchange memory poolExchangeAfter = exchangeProvider.getPoolExchange(exchangeId);
+    assertEq(poolExchangeAfter.reserveBalance, initialReserveBalance + reserveInterest);
+  }
+
+  function test_mintFromInterest_withMaximumInterest() public {
+    uint256 maxInterest = type(uint256).max / poolExchange1.tokenSupply;
+
+    vm.prank(expansionControllerAddress);
+    uint256 amountToMint = exchangeProvider.mintFromInterest(exchangeId, maxInterest);
+
```
philbow61, line 543: Can we maybe add an assertEq here with an expectedAmountToMint that you calculate off-chain and verify against?

test/unit/goodDollar/GoodDollarExchangeProvider.t.sol
RyRy79261: @bowd This is a bit confusing, maximum is the smallest possible scalar, with minimum being 100% - 1/

That reads as the max and min are the same value. What would the maximum be then?

philbow61 (changes requested):
LGTM, left some little nits/suggestions.
Can we change the fuzz tests such that the PoolConfig they are running on is the Config the system will most likely be deployed with. The config is in our G$ discord group chat.

test/unit/goodDollar/GoodDollarExchangeProvider.t.sol
RyRy79261: This was based off not having a way to confirm total minted in a way that isn't redundant

test/unit/goodDollar/GoodDollarExchangeProvider.t.sol
RyRy79261: @philbow61 for the calculation offchain, do you mean dynamically within the test, or hardcoded in the test?

philbow61: just hardcoded in the test 😁

RyRy79261: @philbow61 What values would make the most sense for checkings this?

RyRy79261: Wolfram Alpha isn't providing answers that are easily hardcoded into this: 
https://www.wolframalpha.com/input?i=%28115792089237316195423570985008687907853269984665640564039457584007913129639935+%2F+%287000000000+**+18%29%29++%2F+%28200000+**18%29

The token amount on max interest with expected config, is this fraction:
```23158417847463239084714197001737581570653996933128112807891516801582625927987/85375770842127348531200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
```

test/unit/goodDollar/GoodDollarExchangeProvider.t.sol
```diff
@@ -436,18 +696,133 @@ contract GoodDollarExchangeProviderTest_updateRatioForReward is GoodDollarExchan
...
+    IBancorExchangeProvider.PoolExchange memory poolExchangeAfter = exchangeProvider.getPoolExchange(exchangeId);
+    uint256 priceAfter = exchangeProvider.currentPrice(exchangeId);
+
+    assertEq(
+      poolExchangeAfter.tokenSupply,
+      initialTokenSupply + fuzzedReward,
+      "Token supply should increase by reward amount"
+    );
+    assertEq(poolExchangeAfter.reserveBalance, initialReserveBalance, "Reserve balance should remain unchanged");
+    assertLe(poolExchangeAfter.reserveRatio, initialReserveRatio, "Reserve ratio should decrease");
```
baroooo, line 819:
```suggestion
    assertLe(poolExchangeAfter.reserveRatio, initialReserveRatio, "Reserve ratio should stay the same or decrease");
```

baroooo (approved):

nvtaveras (approved):

PR #530, opened 2024-10-16
GoodDollar Fork Tests

chapati23:
### Description

- [x] Wrote fork tests for Good Dollar contracts
- [x] Wrote fork tests for BancorExchangeProvider (this one's debatable; it's largely the same logic, but I wasn't sure if we might use this contract in isolation in the future)
- [x] Deleted Good Dollar Integration test (this was already a fork test in disguise and all of its logic has been merged into the new good dollar fork tests)
- [x] Added two additional `require`s to `BancorExchangeProvider.executeSwap()` to guard against potential underflows
- [x] Moved `.solhint.test.json` into `./test/.solhint.json` so the VS Code solidity extension can pick it up (rn it's using the main config also for tests which results in lots of annoying red squiggles that are non-issues)
- [x] Removed some Baklava remnants

### How to review

- [ ] Check out this branch: `git checkout chore/gooddollar-fork-tests`
- [ ] Run `yarn fork-test:celo-mainnet`, they should be green (alfajores is broken due to L2 changes unrelated to this PR)
- [ ] Review the code and see if there's anything missing that you think should be covered by the fork tests


baroooo (approved):
contract changes look good, tests will be reviewed together with other changes in the final PR

RyRy79261 (approved):
LGTM, I did have an intermittent test fail on the `rateFeedDependencies_haltDependantTrading` test however I'd assume that's related to state of the chain when forking. 

@chapati23 To confirm if this potential erroring is an issue however I can't deliberately cause the error to happen so I would assume it's a natural behaviour of the on chain logic

PR #532, opened 2024-10-18
test: add BancorExchangeProvider pricing tests

philbow61:
### Description

This Pr adds tests for the pricing logic of the bancor exchange provider.

During the testing, we realized that the BancorFormula.sol contract is missing one function we need.
This function is getAmountIn for the tokenIn being the BancorToken. The new function is called `saleCost()`

The math formula for this function is derived from the `saleTargetAmount()` function already implemented. It's important that the base of power operation is larger one otherwise the power function in that contract reverts. 

These are the current formulas used for the different pricing functions:
image

This is how we derived the formula for `saleCost()`

image

### Related issues

- Fixes https://github.com/mento-protocol/mento-general/issues/453

### How to review 

- verify formula is derived correctly 
- exit contribution is applied correctly 
- tests cover everything

### DIFF
--- a/contracts/goodDollar/BancorFormula.sol
+++ b/contracts/goodDollar/BancorFormula.sol
@@ -559,13 +559,9 @@ contract BancorFormula {
    * - The natural logarithm of the input is calculated by summing up the intermediate results above
    * - For example: log(250) = log(e^4 * e^1 * e^0.5 * 1.021692859) = 4 + 1 + 0.5 + log(1 + 0.021692859)
    */
-  // We're choosing to trust Bancor's audited Math
-  // slither-disable-start divide-before-multiply
   function optimalLog(uint256 x) internal pure returns (uint256) {
     uint256 res = 0;
 
-    // slither false positive, y is initialized as z = y = ...
-    // slither-disable-next-line uninitialized-local
     uint256 y;
     uint256 z;
     uint256 w;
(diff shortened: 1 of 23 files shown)


contracts/goodDollar/BancorFormula.sol
```diff
@@ -297,42 +306,55 @@ contract BancorFormula {
...
     uint8 precision;
-    uint256 baseN = _reserveBalance + _amount;
-    (result, precision) = power(baseN, _reserveBalance, _reserveRatio, MAX_WEIGHT);
-    uint256 temp = (_supply * result) >> precision;
-    return temp - _supply;
+    uint256 baseD = _reserveBalance - _amount;
+    (result, precision) = power(_reserveBalance, baseD, _reserveWeight, MAX_WEIGHT);
+    uint256 temp1 = _supply * result;
+    uint256 temp2 = _supply << precision;
+    return (temp1 - temp2 - 1) / result + 1;
```
RyRy79261, line 357: I'm not entrenched in context enough to assess the potential implications for this implementation. 

But at the minimum, the setting of `baseD`, `temp1` & `temp2` as variables in this function can be removed to improve gas efficiency.  

```
    (result, precision) = power(_reserveBalance, (_reserveBalance - _amount), _reserveWeight, MAX_WEIGHT);
    return ((_supply * result) - (_supply << precision) - 1) / result + 1;
    ```

baroooo (approved):
Great job ✅

RyRy79261 (approved):
Looks good! Only nitpik on the temp variables for gas but likely very low impact

PR #537, opened 2024-10-30
feat: add cCOP fork tests

nvtaveras:
### Description

Add fork tests for $cCOP

### DIFF
--- a/test/fork/BaseForkTest.sol
+++ b/test/fork/BaseForkTest.sol
@@ -129,7 +129,7 @@ abstract contract BaseForkTest is Test {
   }
 
   function transferCeloFromReserve(address to, uint256 amount) internal {
-    vm.prank(address(reserve));
+    vm.prank(address(mentoReserve));
     IERC20(lookup("GoldToken")).transfer(to, amount);
   }
 }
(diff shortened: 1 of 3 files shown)


philbow61 (changes requested):
LGTM.
I added the cCOP exhchage on mainnet but alfajores tests are failing because of 2 reasons:
1. oracle problems 
2. It looks like cCOP proposal wasn't executed on alfajores I assume the reason is that mu0Next was already executed on alfajores and the cCOP upgrade is causing problems with FixidityLib in 0.8

Therefore I would suggest we wait until cCOP is executed on testnets before merging this one

chapati23 (changes requested):
I rebased on top of the latest develop and had to fix some resulting compilation errors to get this to run.

Both alfajores and mainnet fork tests are failing for me rn, alfajores because of what @philbow61 said and 3 rate feeds being down rn

chapati23 (approved):
we fixed the mainnet fork tests now, alfajores is broken right now because of broken rate feeds but we're fine with merging regardless

philbow61 (approved):

PR #542, opened 2024-11-18
Feat: locking l2 upgrade

### LINKED ISSUE #538: Update Locking to handle upcoming block.time change
#### Description
With Celo transitioning to a L2, the block time will be reduced from 5 seconds to 1 second. This change impacts our Locking and Governance mechanisms, which currently depend on a block time of 5 seconds, calculating a "week" as 120960 blocks. With the new block time, a week will be equal 604800 blocks.

To ensure the continuity of functionality after the L2 upgrade, we need to update the WEEK constant and adjust startingPointWeek and roundTimestamp such that the week calculation in the Locking contract remains unchanged post-upgrade.

- Add a NEW_WEEK constant in the contract to accommodate the new block time (1s/block) 604,800 blocks per week.

- Introduce a new function, setL2TransitionBlock, which allows us to specify the exact block number for the L2 transition once it is known. This function, callable only by a Mento Labs-controlled multisig, will:
Calculate the correct newStartingPointWeek.
Register both the L2 transition block number and newStartingPointWeek as variables in the contract.

- The formula for calculating newStartingPointWeek will be:
newStartingPointWeek = oldStartingPointWeek - (setL2TransitionBlock - getEpochShift()) / (NEW_WEEK / 4).

- Given that underflow could occur if the L2 block number becomes too large (approximately 210 days from now), modify roundTimestamp() to handle both positive and negative newStartingPointWeek values:
If newStartingPointWeek is negative, add it to the week calculation and round up to maintain week continuity.

- Update roundTimestamp() to check blockno against the L2 transition block number:
If the current block is greater than or equal to the L2 transition block, calculate the week number using the new constants (NEW_WEEK and newStartingPointWeek).
Otherwise, retain the old implementation with existing constants (startingPointWeek and WEEK).

- Verify getEpochShift() calculations remain consistent, as initial analysis suggests that block number modulus operations remain unaffected by the new block time.

- Conduct forked tests to validate functionality and ensure correct values across a range of block times and scenarios. Tests should cover: totalSupply and pastTotalSupply, balanceOf, Voting power and past voting power, lockedAmount and withdrawableAmount for multiple lock owners

#### Acceptance Criteria:

- [ ] Locking contracts accurately handle both 5s and 1s block times.
- [ ] setL2TransitionBlock function successfully updates and registers L2 block number and newStartingPointWeek.
- [ ] roundTimestamp() correctly adjusts for week calculations post and pre transition.
- [ ] Tests confirm correctness across total supply, balances, voting power, locked and withdrawable amounts for multiple lock owners at various points in time.
- [ ] getEpochShift() consistency is verified using tests.

baroooo:
### Description

Updates Locking Contracts to handle upcoming block.time change

Mainly changes the roundTimestamp() to have a forked implementation that uses different values to calculate week no for the blocks before and after l2 transition. Week numbers that return from roundTimestamp before and after should be continues and consistent.

### Other changes

No

### Tested

Unit and fork tests

### Related issues

- Fixes #538

### DIFF
--- a/contracts/governance/locking/LockingBase.sol
+++ b/contracts/governance/locking/LockingBase.sol
@@ -93,7 +93,7 @@ abstract contract LockingBase is OwnableUpgradeable, IVotesUpgradeable {
    */
   LibBrokenLine.BrokenLine public totalSupplyLine;
 
-  // *************** 
+  // ***************
   // New variables for L2 transition upgrade (4 slots)
   // ***************
   /**
@@ -105,7 +105,7 @@ abstract contract LockingBase is OwnableUpgradeable, IVotesUpgradeable {
    */
   address public mentoLabsMultisig;
   /**
-   * @dev L2 starting point week number used to move the first week after the L2 transition to the last week before the L2 transition
+   * @dev L2 starting point week number
    */
   int256 public l2StartingPointWeek;
   /**
@@ -300,7 +300,7 @@ abstract contract LockingBase is OwnableUpgradeable, IVotesUpgradeable {
    */
   function roundTimestamp(uint32 ts) public view returns (uint32) {
     require(!paused, "locking is paused");
-    
+
     if (ts < getEpochShift()) {
       return 0;
     }
@@ -316,7 +316,7 @@ abstract contract LockingBase is OwnableUpgradeable, IVotesUpgradeable {
 
   /**
    * @notice method returns the amount of blocks to shift locking epoch to on L1 CELO.
-   * we move it to 00-00 UTC Wednesday (approx) by shifting 89964 blocks 
+   * we move it to 00-00 UTC Wednesday (approx) by shifting 89964 blocks
    */
   function getEpochShift() internal view virtual returns (uint32) {
     return 89964;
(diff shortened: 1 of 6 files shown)


baroooo:
I think the ci is failing because they removed regex parsing for version pragma in a recent foundry upgrade which is causing similar problems for other people. It also fails for dev branch at the moment.

It should be fixed once this change is released:
https://github.com/foundry-rs/compilers/pull/223

test/unit/governance/Locking/locking.upgrade.t.sol
```diff
@@ -0,0 +1,362 @@
...
+
+    assertEq(locking.locked(alice), lockedL1W40);
+    assertEq(locking.getAvailableForWithdraw(alice), withdrawableL1W40);
+
+    // week no: 60
+    _incrementBlock(l2Week * 20);
+    assertEq(locking.locked(alice), lockedL1W60);
+    assertEq(locking.getAvailableForWithdraw(alice), withdrawableL1W60);
+  }
+}
```
philbow61, line 362: Nice tests 💯 
Can we maybe add one more verifying locking/delegating works as expected after the L2 changes? 
I don't think it will break but maybe its nice to cover these functions as well 🤔

philbow61 (approved):
LGTM nice work on testing this.
1 thing can we add a test in the fork tests verifying pausing the locking contract results in 
in Governance operations are being blocked:
- propose 
- voting

contracts/governance/locking/LockingBase.sol
```diff
@@ -256,20 +312,29 @@ abstract contract LockingBase is OwnableUpgradeable, IVotesUpgradeable {
 
   /**
    * @notice Calculates the week number for a given blocknumber
+   * @dev It takes L2 transition into account to calculate the week number consistently
    * @param ts block number
    * @return week number the block number belongs to
    */
   function roundTimestamp(uint32 ts) public view returns (uint32) {
```
bowd, line 319: I wonder if, given we're touching this code anyway, if we should rename `ts` to `blockNumber` because this `ts` still trips me up when I read it. Is there a broader reason why this is called `ts`?

bowd (changes requested):
Great work 🙏 😍 - have only cosmetic comments, and more meant as a discussion, so let me know your thoughts.

contracts/governance/locking/LockingBase.sol
```diff
@@ -256,20 +312,29 @@ abstract contract LockingBase is OwnableUpgradeable, IVotesUpgradeable {
...
+    require(!paused, "locking is paused");
+
     if (ts < getEpochShift()) {
       return 0;
     }
-    uint32 shifted = ts - (getEpochShift());
-    return shifted / WEEK - uint32(startingPointWeek);
+
+    if (l2Block == 0 || ts < l2Block) {
+      uint32 shifted = ts - getEpochShift();
```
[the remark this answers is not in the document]
baroooo: I think this is done to mock a constant value in tests but I agree it is ugly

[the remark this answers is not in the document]
baroooo: created L1_EPOCH_SHIFT and updated getEpochShift to work with two epoch shifts 👍

bowd (approved):
Love it, way more readable now!

contracts/governance/locking/LockingBase.sol
```diff
@@ -84,6 +92,31 @@ abstract contract LockingBase is OwnableUpgradeable, IVotesUpgradeable {
    * @dev Total supply line of veMento
    */
   LibBrokenLine.BrokenLine public totalSupplyLine;
+
+  // ***************
+  // New variables for L2 transition upgrade (3 slots)
+  // ***************
+  /**
+   * @dev L2 transtion block number
```
chapati23, line 100:
```suggestion
   * @dev L2 transition block number
```

contracts/governance/locking/LockingBase.sol
```diff
@@ -348,9 +429,59 @@ abstract contract LockingBase is OwnableUpgradeable, IVotesUpgradeable {
...
   }
+  /**
+   * @notice Sets the Mento Labs multisig address
+   * @param mentoLabsMultisig_ address of the Mento Labs multisig
+   *
+   */
+  function setMentoLabsMultisig(address mentoLabsMultisig_) external onlyOwner {
+    mentoLabsMultisig = mentoLabsMultisig_;
+    emit SetMentoLabsMultisig(mentoLabsMultisig_);
+  }
```
chapati23, line 443:
```suggestion
  }
 
```

contracts/governance/locking/LockingBase.sol
```diff
@@ -348,9 +429,59 @@ abstract contract LockingBase is OwnableUpgradeable, IVotesUpgradeable {
...
+  /**
+   * @notice Sets the L2 transition block number and pauses locking and governance
+   * @param l2TransitionBlock_ block number of the L2 transition
+   *
```
chapati23, line 447:
```suggestion
```

contracts/governance/locking/LockingBase.sol
```diff
@@ -348,9 +429,59 @@ abstract contract LockingBase is OwnableUpgradeable, IVotesUpgradeable {
...
+   */
```
chapati23, line 448:
```suggestion
   */

```

PR #546, opened 2024-11-21
fix: Removed redundant require

RyRy79261:
### Description

This PR removes a redundant require check which is already present in the next step of the function

### Tested

`forge test` 

### Related issues

- Fixes [#603](https://github.com/mento-protocol/mento-general/issues/603)


bayological (approved):

philbow61 (approved):
LGTM

PR #550, opened 2024-11-22
fix: ensure 0 deltaFlow doesn't update state

philbow61:
### Description

This Pr fixes in issue in the way our trading limits handle subunit amounts. Before when :
`deltaFlowUnits == 0 `
`deltaFlowUnits `was set to either 1 or -1 without checking whether the deltaFlow is 0. 
This resulted in a swap with 0 amounts updating the trading limit state  by -1.

### Tested

- added new test ensuring sate is not updated on 0 deltaFlow

### Related issues

- https://github.com/mento-protocol/mento-general/issues/601

### DIFF
--- a/test/unit/libraries/TradingLimits.t.sol
+++ b/test/unit/libraries/TradingLimits.t.sol
@@ -261,9 +261,9 @@ contract TradingLimitsTest is Test {
 
   function test_update_withZeroDeltaFlow_doesNotUpdate() public {
     state = harness.update(state, configL0L1LG(300, 1000, 1 days, 10000, 1000000), 0, 18);
-    assertEq(state.netflow0, state.netflow0);
-    assertEq(state.netflow1, state.netflow1);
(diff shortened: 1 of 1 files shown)


RyRy79261 (approved):

test/unit/libraries/TradingLimits.t.sol
```diff
@@ -259,6 +259,13 @@ contract TradingLimitsTest is Test {
     assertEq(state.netflowGlobal, 0);
   }
 
+  function test_update_withZeroDeltaFlow_doesNotUpdate() public {
+    state = harness.update(state, configL0L1LG(300, 1000, 1 days, 10000, 1000000), 0, 18);
```
baroooo, line 263: the state var is being shadowed here making this test effectively a tautology

baroooo (changes requested):

PR #553, opened 2024-11-22
fix: setting of last expansion timestamp

philbow61:
### Description

fixes issue in how lastUpdated was set:
before if  3.5 days passed and a daily expansion was configured, 3 expansions were calculated and the last expansion timestamp was updated to now essentially losing half a day. Over time this could have resulted in the yearly expansion rate not being achieved. 

### Tested

-  added tests for this behavior 

### Related issues

https://github.com/mento-protocol/mento-general/issues/592


RyRy79261 (approved):
LGTM

baroooo (approved):

PR #554, opened 2024-11-22
Feat: slippage protection for reserve ratio

baroooo:
### Description

Adds slippage protection to `updateRatioForReward()`

### Other changes

Simplifies the new ratio calculation formula

### Tested

Unit tests

### Related issues

- Fixes [#594](https://github.com/mento-protocol/mento-general/issues/594)

### DIFF
--- a/contracts/goodDollar/GoodDollarExchangeProvider.sol
+++ b/contracts/goodDollar/GoodDollarExchangeProvider.sol
@@ -241,9 +241,4 @@ contract GoodDollarExchangeProvider is IGoodDollarExchangeProvider, BancorExchan
    * See https://docs.openzeppelin.com/contracts/4.x/upgradeable#storage_gaps
    */
   uint256[50] private __gap;
-
-  // Helper function to calculate absolute value
-  function abs(int256 x) internal pure returns (uint256) {
-    return x < 0 ? uint256(-x) : uint256(x);
-  }
 }
--- a/contracts/goodDollar/GoodDollarExchangeProvider.sol
+++ b/contracts/goodDollar/GoodDollarExchangeProvider.sol
@@ -192,7 +192,11 @@ contract GoodDollarExchangeProvider is IGoodDollarExchangeProvider, BancorExchan
    * @dev Calculates the new reserve ratio needed to mint the G$ reward while keeping the current price the same.
    *      calculation: newRatio = (tokenSupply * reserveRatio) / (tokenSupply + reward)
    */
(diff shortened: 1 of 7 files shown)


contracts/goodDollar/GoodDollarExchangeProvider.sol
```diff
@@ -190,21 +190,31 @@ contract GoodDollarExchangeProvider is IGoodDollarExchangeProvider, BancorExchan
...
+    uint256 newScaledRatio = unwrap(numerator.div(denominator));
+
+    uint32 newRatioUint = uint32(newScaledRatio / 1e10);
+
+    require(newRatioUint > 0, "New ratio must be greater than 0");
 
-    UD60x18 numerator = wrap(exchange.reserveBalance);
-    UD60x18 denominator = wrap(exchange.tokenSupply + rewardScaled).mul(wrap(currentPriceScaled));
-    uint256 newRatioScaled = unwrap(numerator.div(denominator));
+    uint256 allowedSlippage = (exchange.reserveRatio * maxSlippagePercentage) / 100;
```
baroooo, line 213: is percentage good enough? should we use BPS instead to make it more flexible?

contracts/goodDollar/GoodDollarExpansionController.sol
```diff
@@ -190,12 +190,24 @@ contract GoodDollarExpansionController is IGoodDollarExpansionController, Pausab
...
+  /// @inheritdoc IGoodDollarExpansionController
+  function mintRewardFromReserveRatio(
+    bytes32 exchangeId,
+    address to,
+    uint256 amount,
+    uint256 maxSlippagePercentage
+  ) public onlyAvatar {
     require(to != address(0), "Recipient address must be set");
     require(amount > 0, "Amount must be greater than 0");
+    require(maxSlippagePercentage <= 100, "Max slippage percentage cannot be greater than 100%");
```
philbow61, line 206: should we check for 0 as well. For super small rewards the call could still go through but generally speaking, the ratio should change looking at the function name in GoodDollarExchangeProvider 😁

contracts/goodDollar/GoodDollarExchangeProvider.sol
philbow61: yes why not we could use maxWeight(1e8) as 100%

philbow61 (approved):
LGTM left 2 samll comments ✌️

contracts/goodDollar/GoodDollarExpansionController.sol
philbow61: haha true 🤝

PR #556, opened 2024-11-25
Fix/use correct min for int48

baroooo:
### Description

Fixes Sherlock issue 46 by using `type(int48).min`

### Other changes

updates test for withOverflowOnAdd to test the correct limit

### Tested

unit tests

### Related issues

- Fixes [#598](https://github.com/mento-protocol/mento-general/issues/598)

### DIFF
--- a/test/unit/libraries/TradingLimits.t.sol
+++ b/test/unit/libraries/TradingLimits.t.sol
@@ -309,10 +309,23 @@ contract TradingLimitsTest is Test {
 
   function test_update_withOverflowOnAdd_reverts() public {
     ITradingLimits.Config memory config = configLG(int48(uint48(2 ** 47)));
-    int256 maxFlow = int256(uint256(type(uint48).max / 2));
+    int256 maxFlow = int256(type(int48).max);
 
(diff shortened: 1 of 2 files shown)


philbow61 (approved):
LGTM left one small nit

test/unit/libraries/TradingLimits.t.sol
```diff
@@ -309,10 +309,23 @@ contract TradingLimitsTest is Test {
...
 
     state = harness.update(state, config, (maxFlow - 1000) * 1e18, 18);
+    state = harness.update(state, config, 1000 * 1e18, 18);
+
+    vm.expectRevert(bytes("int48 addition overflow"));
+    state = harness.update(state, config, 1 * 1e18, 18);
+  }
+
+  function test_update_withUnderflowOnAdd_reverts() public {
+    ITradingLimits.Config memory config = configLG(int48(uint48(2 ** 47)));
```
philbow61, line 322: nit: this results in the global limit being a negative value which is a state we are not supporting when setting trading Limits

bayological (approved):

Issue #555, opened 2024-11-22
G$: Issue #64


philbow61:
**Descirption:** 
This Pr is about fixing the discontinued curve problems on taking the exit contribution

PR #557, opened 2024-11-25
Test/new exit contribution

philbow61:
### Description

This PR tries to solve the discontinued curve problem that was causing the same sell amount when executed in two swaps resulting in a larger amountOut than when sold in one swap. 
The idea is to apply the exitContribution on the G$ amount being sold and burning that amount. In order to not move on the curve by this burn we increase the reserve ratio in order to keep the price before the exitContribution burn equal to the price after. 

### Other changes

Added requires that ensure the token Supply always stays at a minimum of 1 wei

### Tested

- new tests added ensuring the difference in splitting sells into multiple swaps stays super small compared to the gas costs   

### Related issues

 https://github.com/mento-protocol/mento-general/issues/596
https://github.com/mento-protocol/mento-core/issues/555


baroooo (approved):
committed my suggestions instead of adding review comments

https://github.com/mento-protocol/mento-core/pull/557/commits/2fc70ce50480253b69add4bff9a7c3a545e8af15

bowd (approved):
No comments, immaculate work 😂

baroooo (approved):

PR #558, opened 2024-11-25
Fix: Exchange created with zero reserve

RyRy79261:
### Description

This PR adds a check to prevent exchanges with zero reserve being created.

This rendered 4 tests redundant due to an exchange of this nature not being possible at creation

### Related issues

- Fixes [#602](https://github.com/mento-protocol/mento-general/issues/602)

### DIFF
--- a/test/unit/goodDollar/BancorExchangeProvider.t.sol
+++ b/test/unit/goodDollar/BancorExchangeProvider.t.sol
@@ -164,6 +164,12 @@ contract BancorExchangeProviderTest_initilizerSettersGetters is BancorExchangePr
     bancorExchangeProvider.setExitContribution(exchangeId, 1e5);
   }
 
+  function test_setExitContribution_whenReserveBalanceIsZero_shouldRevert() public {
+    poolExchange1.reserveBalance = 0;
+    vm.expectRevert("Reserve balance must be greater than 0");
+    bancorExchangeProvider.createExchange(poolExchange1);
+  }
+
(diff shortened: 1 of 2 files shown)


baroooo (approved):

test/unit/goodDollar/BancorExchangeProvider.t.sol
```diff
@@ -453,10 +453,20 @@ contract BancorExchangeProviderTest_getAmountIn is BancorExchangeProviderTest {
...
+      reserveRatio: 1e8 * 0.2,
+      exitContribution: 0
+    });
+    bytes32 exchangeId = bancorExchangeProvider.createExchange(poolExchangeTemp);
+
+    vm.prank(brokerAddress);
+    bancorExchangeProvider.swapOut(exchangeId, address(token), address(reserveToken), 60_000 * 1e18);
 
-    vm.expectRevert("ERR_INVALID_RESERVE_BALANCE");
+    vm.expectRevert("ERR_INVALID_SUPPLY");
```
nvtaveras, line 469: Since these tests are no longer valid we should get rid of them (specially because the name of the test doesn't follow the error message) and write a test that tries to create an exchange with a 0 reserve balance.

RyRy79261: So was chatting to @baroooo who wanted to keep them in for regression tests, I initially removed them, but then added them back, and due to the supply check happening first, this error is throw first

nvtaveras (approved):

PR #572, opened 2024-12-19
feat: add GD proxies

### LINKED ISSUE #566: G$ - deployment script

philbow61:
### Description

- This PR adds the named proxy contracts for the upcoming G$ deployment. 
- We are using the OZ TransparentUpgradeableProxy contract for the exchangeProvider as well as the expansionController
- For the GoodDollarReserve proxy we went with the same celo proxy implementation used by the existing mento reserve proxy. The Reason for not using the newer more common OZ proxy implementations is to not make future reserve upgrades too complicated and have different proxy logics point to the same implementation contract. 

If anybody is strongly in favor of using the newer OZ proxy implementation or has some additional thoughts on this lmk.

### Related issues

- Fixes #566


bayological (approved):
LGTM.

Issue #10, opened 2022-09-22
McMint


bowd:
### Description

MultiCollateral-Mento or “McMint” is a generalization of the current Mento system to allow virtual pools between any mento stable asset and any mento collateral asset to be traded under different exchange mechanisms (constant-product, constant-sum, etc.). In contrast to the current Mento, which only permits trading against CELO, and with a constant-product mechanism.

### Motivation
The introduction of McMint will allow the minting of large amounts of stable assets with low slippage. This is currently not possible with the current version of Mento. McMint will also make it possible to deploy new stables with minimal deployments and makes it possible to trade stable to stable.   

### Definition of Done

- The Mento protocol allows stable assets to be minted/burned using any collateral asset held by the reserve
- The Mento protocol allows the creation of "virtual pools" between any Mento stable & collateral asset using different exchange mechanisms

### Linked Stories

- #44 
- #32 
- #53 
- #30 
- #43 
- #31 
- #17 
- #29 
- #33 
- #28 
- #60 
- #24  
- #54 
- #55

PR #575, opened 2025-01-29
Fix: Relocking min amount bypass

RyRy79261:
### Description

[Hats issue #10](https://github.com/hats-finance/Mento-0x2a1b9b1f6fa7c2e73815a7dff0e1688767382694/issues/10) identified that in the event of relocking a completed lock, there was no minimum check that was present in a standard lock, this was given the assumption that relocking would only occur on non expired locks.

### DIFF
--- a/contracts/governance/locking/LockingRelock.sol
+++ b/contracts/governance/locking/LockingRelock.sol
@@ -68,6 +68,7 @@ abstract contract LockingRelock is LockingBase {
     uint32 toTime
   ) internal view {
     require(newAmount > 0, "zero amount");
+    require(newAmount >= 1e18, "amount is less than minimum");
     require(newCliff <= MAX_CLIFF_PERIOD, "cliff too big");
     require(newSlopePeriod <= MAX_SLOPE_PERIOD, "slope period too big");
     require(newSlopePeriod > 0, "slope period equal 0");
@@ -117,7 +118,6 @@ abstract contract LockingRelock is LockingBase {
    */
   function rebalance(uint256 id, address account, uint96 bias, uint96 residue, uint96 newAmount) internal {
     require(residue <= newAmount, "Impossible to relock: less amount, then now is");
-    require(newAmount >= 1e18, "amount is less than minimum");
     uint96 addAmount = newAmount - (residue);
     uint96 amount = accounts[account].amount;
     uint96 balance = amount - (bias);
--- a/contracts/governance/locking/LockingRelock.sol
+++ b/contracts/governance/locking/LockingRelock.sol
@@ -117,6 +117,7 @@ abstract contract LockingRelock is LockingBase {
    */
   function rebalance(uint256 id, address account, uint96 bias, uint96 residue, uint96 newAmount) internal {
     require(residue <= newAmount, "Impossible to relock: less amount, then now is");
+    require(newAmount >= 1e18, "amount is less than minimum");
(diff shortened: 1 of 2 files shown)


contracts/governance/locking/LockingRelock.sol
```diff
@@ -117,6 +117,7 @@ abstract contract LockingRelock is LockingBase {
    */
   function rebalance(uint256 id, address account, uint96 bias, uint96 residue, uint96 newAmount) internal {
     require(residue <= newAmount, "Impossible to relock: less amount, then now is");
+    require(newAmount >= 1e18, "amount is less than minimum");
```
philbow61, line 120: Why did you move this check into the rebalance function? I would say it better fits into the verification() function that is called before rebalance() and already performs a check on newAmount 🤔 
We could also think about adding the check on top of relock() to fail early and be more consistent with lock().

philbow61 (changes requested):
LGTM left one comment 👍

contracts/governance/locking/LockingRelock.sol
RyRy79261: Makes sense, will shuffle around 👌🏼

contracts/governance/locking/LockingRelock.sol
```diff
@@ -68,6 +68,7 @@ abstract contract LockingRelock is LockingBase {
     uint32 toTime
   ) internal view {
     require(newAmount > 0, "zero amount");
+    require(newAmount >= 1e18, "amount is less than minimum");
```
philbow61, line 71: having both checks here is redundant

contracts/governance/locking/LockingRelock.sol
```diff
@@ -67,7 +67,7 @@ abstract contract LockingRelock is LockingBase {
     uint32 newCliff,
     uint32 toTime
   ) internal view {
-    require(newAmount > 0, "zero amount");
+    require(newAmount >= 1e18, "amount is less than minimum");
```
nvtaveras, line 70: I agree with what @philbow61 said in another comment that it would be nice to put the check at the top of relock() for consistency with the lock function https://github.com/mento-protocol/mento-core/blob/develop/contracts/governance/locking/Locking.sol#L54

We also already have a delegate address check inside of the relock function already. wdyt?

test/unit/governance/Locking/relock.t.sol
```diff
@@ -646,4 +646,33 @@ contract Relock_LockingTest is LockingTest {
...
+  function test_relock_whenAmountIsMinimum_shouldRelockSuccessfully() public {
+    mentoToken.mint(alice, 100e18);
+
+    vm.prank(alice);
+    lockId = locking.lock(alice, alice, 30e18, 3, 3);
+
+    // Wait until lock expires so residue is 0
+    _incrementBlock(6 * weekInBlocks);
+
+    // Should succeed with amount = minimum
```
nvtaveras, line 674: nit: the comments in here are redundant given the names of the test, same for the other one

contracts/governance/locking/LockingRelock.sol
RyRy79261: So it is in the verification, which seems more semantically correct, so it's down to what makes more consistent sense, I don't have a strong preference for either

nvtaveras (approved):

PR #576, opened 2025-02-06
Feat: Disable Initializer

RyRy79261:
### Description

This PR addresses the implementation having an exposed and callable, however without any structural risk, initializer in future implementations of the Locking contract

### DIFF
--- a/contracts/governance/deployers/LockingDeployerLib.sol
+++ b/contracts/governance/deployers/LockingDeployerLib.sol
@@ -10,6 +10,6 @@ library LockingDeployerLib {
    * @return The address of the new Locking contract
    */
   function deploy() external returns (Locking) {
-    return new Locking();
+    return new Locking(false);
   }
 }
--- a/contracts/governance/deployers/LockingDeployerLib.sol
+++ b/contracts/governance/deployers/LockingDeployerLib.sol
@@ -10,6 +10,6 @@ library LockingDeployerLib {
    * @return The address of the new Locking contract
    */
   function deploy() external returns (Locking) {
-    return new Locking(false);
+    return new Locking(true);
   }
 }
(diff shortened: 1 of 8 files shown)


contracts/governance/deployers/LockingDeployerLib.sol
```diff
@@ -10,6 +10,6 @@ library LockingDeployerLib {
    * @return The address of the new Locking contract
    */
   function deploy() external returns (Locking) {
-    return new Locking();
+    return new Locking(false);
```
nvtaveras, line 13: I don't think we are using this lib again but shouldn't it default to true? See https://github.com/mento-protocol/mento-core/blob/dc5eec06d16c12e778b294307e0f09a9124e05de/contracts/governance/deployers/EmissionDeployerLib.sol#L13

RyRy79261: The constructor accepts a boolean regarding disabling the initializers

test/fork/upgrades/LockingUpgradeForkTest.sol
```diff
@@ -59,7 +59,7 @@ contract LockingUpgradeForkTest is BaseForkTest {
     mentoGovernor = governanceFactory.mentoGovernor();
     mentoToken = governanceFactory.mentoToken();
 
-    newLockingImplementation = address(new Locking());
+    newLockingImplementation = address(new Locking(false));
```
baroooo, line 62: this should be called with true since we are simulating the actual upgrade on a fork. it should be the same as the real world upgrade

contracts/governance/deployers/LockingDeployerLib.sol
baroooo: I agree with Nelson, although the lib won't be used again technically it is an implementation behind the proxy and initializers should be disabled.

baroooo (changes requested):

contracts/governance/deployers/LockingDeployerLib.sol
RyRy79261: Change restricted only to the lib file, push added

nvtaveras (approved):

baroooo (approved):

PR #580, opened 2025-03-11
feat: add proxies for GBP, JPY, NGN and ZAR

nvtaveras:
### Description

Add the StableTokenXXXProxy contract for GBP, JPY, NGN and ZAR


chapati23 (approved):
let's go

chapati23 (approved):
go

PR #581, opened 2025-03-11
Release v2.6.3

nvtaveras:
### Description

New release after adding the new StableTokenXXX proxy contracts so that we can use them in mento-deployment https://github.com/mento-protocol/mento-core/pull/580

It seems like I messed up the release when I did v2.6.2 since the diff shows some commits that should have been merged as part of https://github.com/mento-protocol/mento-core/pull/578. Actual diff should only be https://github.com/mento-protocol/mento-core/commit/0ca06764811c6851d5cc276dc888a811a6a6b617


chapati23 (approved):
lgtm, did a quick sanity check through the critical functions and all looks good

baroooo (approved):

PR #595, opened 2025-05-12
Feat/fpmm

baroooo:
### Description

This PR introduces the Fixed Price Market Maker contract, a core component of the Mento v3 protocol that facilitates token swaps at fixed exchange rates. The FPMM manages a liquidity pool of two tokens and enables trading between them based on oracle rates. It implements a market maker model similar to Uniswap and uses an interface similar to Velodrome but with pricing based on external oracles and with a flash-loan-like rebalancing mechanism for various liquidity strategies.

Key features of the FPMM implementation:
- Oracle-based pricing for token swaps using the SortedOracles contract. This is subject to change and we might swap SorterOracles with another implementation.
- Support for rebalancing mechanism to realign pool reserves with oracle prices
- Circuit breaker integration via BreakerBox contract
- Liquidity provision mechanism with ERC20 LP tokens
- Fee management with configurable protocol fees
- Decimal-aware calculations supporting tokens with different decimal places

### Tested
The implementation is covered by unit and integration tests.
Fuzz tests will be added in a separate PR


contracts/swap/FPMM.sol
```diff
@@ -0,0 +1,463 @@
...
+    uint256 amount,
+    uint256 fromDecimals,
+    uint256 toDecimals,
+    uint256 numerator,
+    uint256 denominator
+  ) public pure returns (uint256) {
+    if (fromDecimals > toDecimals) {
+      return (amount * numerator) / ((denominator * fromDecimals) / toDecimals);
+    } else if (fromDecimals < toDecimals) {
+      return ((amount * numerator * toDecimals) / fromDecimals) / denominator;
```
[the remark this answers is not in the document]
baroooo: yeah I had it like that then changed it because of slither since I did not want to ignore a slither warning when not needed after seeing the worst case scenario was not realistic:
```
fromDecimals: 1e17
toDecimals: 1e18
numerator: 1_000_000e18

in order to overflow, amount needs to be = 115792089237316195423570985008687907 which is equal to 1.157.920.892.373.161.954 units
```
but taking another look, I think the function is harder to read now and adding the slither ignore seems like a better option

contracts/swap/FPMM.sol
```diff
@@ -0,0 +1,463 @@
...
+  }
+
+  /// @inheritdoc IFPMM
+  function swap(uint256 amount0Out, uint256 amount1Out, address to, bytes calldata data) external nonReentrant {
+    _swap(amount0Out, amount1Out, to, data, false);
+  }
+
+  /// @inheritdoc IFPMM
+  function rebalance(uint256 amount0Out, uint256 amount1Out, address to, bytes calldata data) external nonReentrant {
+    require(liquidityStrategy[msg.sender], "FPMM: NOT_LIQUIDITY_STRATEGY");
```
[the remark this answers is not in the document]
baroooo: I will have both checks since I can not see a use case where another address calling rebalance with `to` being a liquidity strategy.

contracts/swap/FPMM.sol
```diff
@@ -0,0 +1,463 @@
...
+    SwapData memory swapData;
+
+    (swapData.rateNumerator, swapData.rateDenominator) = sortedOracles.medianRate(referenceRateFeedID);
+    swapData.initialReserveValue = totalValueInToken1(
+      _reserve0,
+      _reserve1,
+      swapData.rateNumerator,
+      swapData.rateDenominator
+    );
+    swapData.initialPriceDifference = _calculatePriceDifference();
```
philbow61, line 394: the _calculatePriceDifference() function calls getPrices(), which again gets the medianRate and the reserves. Since we already got/read those values in _swap, we could create an internal function which only does the calculation and is called by both the public facing function and _swap, so we don't have to read/call twice and save some gas

contracts/swap/FPMM.sol
```diff
@@ -0,0 +1,463 @@
...
+  function _rebalanceCheck(SwapData memory swapData) internal view returns (uint256 newPriceDifference) {
+    uint256 newReserveValue = totalValueInToken1(reserve0, reserve1, swapData.rateNumerator, swapData.rateDenominator);
+    newPriceDifference = _calculatePriceDifference();
+
+    // Ensure price difference is smaller than before
+    require(newPriceDifference < swapData.initialPriceDifference, "FPMM: PRICE_DIFFERENCE_NOT_IMPROVED");
+    require(newPriceDifference < rebalanceThreshold, "FPMM: POOL_NOT_REBALANCED");
+
+    // Check for excessive value loss
+    uint256 rebalanceIncentiveAmount = (swapData.initialReserveValue * rebalanceIncentive) / 10000;
```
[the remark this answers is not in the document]
philbow61: Yeah I agree. I think there are some aspects we need to think about more:
- balance between flashloan fee vs rebalance incentive 
- maybe we can also work around a min<percentage<max incentive

contracts/swap/FPMM.sol
```diff
@@ -0,0 +1,468 @@
...
+    reservePrice = (reserve1 * _decimals0 * 1e18) / (reserve0 * _decimals1);
+  }
+
+  /// @inheritdoc IFPMM
+  function getAmountOut(uint256 amountIn, address tokenIn) public view returns (uint256 amountOut) {
+    require(tokenIn == token0 || tokenIn == token1, "FPMM: INVALID_TOKEN");
+
+    if (amountIn == 0) return 0;
+
+    (uint256 rateNumerator, uint256 rateDenominator) = sortedOracles.medianRate(referenceRateFeedID);
```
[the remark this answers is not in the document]
baroooo: two main reasons:
- I wanted the convert function to be more general
- by doing so we reduce the calls to the oracle because convert is called multiple times in a swap

contracts/swap/FPMM.sol
baroooo: yeah I had it like that at first but then I needed to _calculatePriceDifference once with the initial values and then get the new reserve and call again second time in the _swap to calculate the new ratio + handle extra variables in scope which caused stack too deep. I agree with you but the other option also becomes ugly. lmk if you come up with a better way of doing it

contracts/swap/FPMM.sol
```diff
@@ -0,0 +1,478 @@
...
+
+  /// @inheritdoc IFPMM
+  function setLiquidityStrategy(address strategy, bool state) external onlyOwner {
+    liquidityStrategy[strategy] = state;
+    emit LiquidityStrategyUpdated(strategy, state);
+  }
+
+  /// @inheritdoc IFPMM
+  function setSortedOracles(address _sortedOracles) public onlyOwner {
+    require(_sortedOracles != address(0), "SortedOracles address must be set");
```
bayological, line 305: Let's get these error messages in the same format with the "FPMM:" prefix

philbow61 (changes requested):

contracts/swap/FPMM.sol
```diff
@@ -0,0 +1,657 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+pragma solidity 0.8.18;
+
+import { IFPMM } from "../interfaces/IFPMM.sol";
+import { ERC20Upgradeable } from "openzeppelin-contracts-upgradeable/contracts/token/ERC20/ERC20Upgradeable.sol";
+import { OwnableUpgradeable } from "openzeppelin-contracts-upgradeable/contracts/access/OwnableUpgradeable.sol";
+import { ReentrancyGuard } from "openzeppelin-contracts-next/contracts/security/ReentrancyGuard.sol";
```
bayological, line 7: nit: any reason for not importing from the same version of oz?

bayological (approved):

test/unit/swap/FPMM/FPMMMint.t.sol
```diff
@@ -0,0 +1,245 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// solhint-disable var-name-mixedcase
+pragma solidity ^0.8;
+
+import { FPMMBaseTest } from "./FPMMBaseTest.sol";
+import { IERC20 } from "openzeppelin-contracts-next/contracts/token/ERC20/IERC20.sol";
+
+contract FPMMMintTest is FPMMBaseTest {
```
philbow61, line 8: nit: amybe add an expect emit of the MInt event in one of the tests to ensure the correct values are emitted

test/unit/swap/FPMM/FPMMBurn.t.sol
```diff
@@ -0,0 +1,200 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// solhint-disable var-name-mixedcase
+pragma solidity ^0.8;
+
```
philbow61, line 4: nit: maybe add one emit check of the burn event

contracts/swap/FPMM.sol
```diff
@@ -0,0 +1,657 @@
...
+
+  /// @inheritdoc IFPMM
+  function getAmountOut(uint256 amountIn, address tokenIn) public view returns (uint256 amountOut) {
+    FPMMStorage storage $ = _getFPMMStorage();
+
+    require(tokenIn == $.token0 || tokenIn == $.token1, "FPMM: INVALID_TOKEN");
+
+    if (amountIn == 0) return 0;
+
+    (uint256 rateNumerator, uint256 rateDenominator) = $.sortedOracles.medianRate($.referenceRateFeedID);
```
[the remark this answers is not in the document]
philbow61: regarding this i think we should also check the trading mode here to not provide a quote that can't be traded

test/unit/swap/FPMM/FPMMGetAmountOut.t.sol
```diff
@@ -0,0 +1,128 @@
...
+    address invalidToken = makeAddr("INVALID_TOKEN");
+
+    vm.expectRevert("FPMM: INVALID_TOKEN");
+    fpmm.getAmountOut(100, invalidToken);
+  }
+
+  function test_getAmountOut_whenRateIsOneToOne_shouldReturnCorrectAmount()
+    public
+    initializeFPMM_withDecimalTokens(18, 18)
+    setupMockOracleRate(10e18, 100e18)
```
philbow61, line 26: This is a bit contradictory to the test name, right?

contracts/swap/FPMM.sol
```diff
@@ -0,0 +1,657 @@
...
+    uint256 newReserveValue = _totalValueInToken1(
+      $.reserve0,
+      $.reserve1,
+      swapData.rateNumerator,
+      swapData.rateDenominator
+    );
+
+    uint256 fee0 = (swapData.amount0In * $.protocolFee) / BASIS_POINTS_DENOMINATOR;
+    uint256 fee1 = (swapData.amount1In * $.protocolFee) / BASIS_POINTS_DENOMINATOR;
+
```
philbow61, line 643: I think this fee calculation behaves a bit differently than what we want to achieve:
Here we calculate the fee on the amountIn, which already includes the fee.
I would suggest we calculate the expected fee here based on the amount out taken.

contracts/swap/FPMM.sol
baroooo: for some reason upgradable version of the safeERC20 does not work and I could not figure out why. I changed others to use the upgradable versions. let me know if you can figure out what would be the reason

contracts/swap/FPMM.sol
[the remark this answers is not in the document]
baroooo: validity checks are not there since this is more like a placeholder and we will replace this with another oracle implementation.

added the trading mode check to the getAmountOut

contracts/swap/FPMM.sol
```diff
@@ -616,25 +609,22 @@ contract FPMM is IFPMM, ReentrancyGuardUpgradeable, ERC20Upgradeable, OwnableUpg
...
+          swapData.amount0Out,
+          $.decimals0,
+          $.decimals1,
+          swapData.rateNumerator,
+          swapData.rateDenominator
+        ) * $.rebalanceIncentive) /
+        BASIS_POINTS_DENOMINATOR;
     } else {
-      swapValue = swapData.amount1In;
+      rebalanceIncentiveInToken1 = (swapData.amount1Out * $.rebalanceIncentive) / BASIS_POINTS_DENOMINATOR;
```
baroooo, line 626: nit: I would still prefer getting the swapped amount in token 1 and applying incentive outside of the if for better readability instead of doing `amount * incentive /BPS` twice

contracts/swap/FPMM.sol
```diff
@@ -616,25 +609,22 @@ contract FPMM is IFPMM, ReentrancyGuardUpgradeable, ERC20Upgradeable, OwnableUpg
...
-      uint256 amount0InValue = convertWithRate(
-        swapData.amount0In,
-        $.decimals0,
-        $.decimals1,
-        swapData.rateNumerator,
-        swapData.rateDenominator
-      );
-      swapValue = amount0InValue;
+    uint256 rebalanceIncentiveInToken1;
+    if (swapData.amount0Out > 0) {
```
baroooo, line 615: what was the reason to apply the incentive on amountOut?

contracts/swap/FPMM.sol
```diff
@@ -406,27 +409,24 @@ contract FPMM is IFPMM, ReentrancyGuardUpgradeable, ERC20Upgradeable, OwnableUpg
...
-    swapData.amount1In = swapData.balance1 > $.reserve1 - amount1Out
-      ? swapData.balance1 - ($.reserve1 - amount1Out)
-      : 0;
-    require(swapData.amount0In > 0 || swapData.amount1In > 0, "FPMM: INSUFFICIENT_INPUT_AMOUNT");
+    uint256 amount0In = balance0 > $.reserve0 - amount0Out ? balance0 - ($.reserve0 - amount0Out) : 0;
+    uint256 amount1In = balance1 > $.reserve1 - amount1Out ? balance1 - ($.reserve1 - amount1Out) : 0;
+
+    require(
+      (amount0Out > 0 && amount1In > 0 && amount0In == 0) || (amount1Out > 0 && amount0In > 0 && amount1In == 0),
+      "FPMM: INSUFFICIENT_INPUT_AMOUNT"
```
baroooo, line 429: we should come up with a better error message here, this might be confusing as it covers more cases now. 
maybe something like: REBALANCE_DIRECTION_INVALID

contracts/swap/FPMM.sol
```diff
@@ -0,0 +1,682 @@
...
+    uint256 toDecimals,
+    uint256 numerator,
+    uint256 denominator
+  ) public pure returns (uint256) {
+    if (fromDecimals > toDecimals) {
+      uint256 decimalAdjustment = fromDecimals / toDecimals;
+      return (amount * numerator) / (denominator * decimalAdjustment);
+    } else if (fromDecimals < toDecimals) {
+      uint256 decimalAdjustment = toDecimals / fromDecimals;
+      return (amount * numerator * decimalAdjustment) / denominator;
```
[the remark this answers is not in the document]
baroooo: a pessimistic yet logical worst case:
oracle rate = 1_000_000 e18
adjustment = e12

 amount needs to be 115792089237316195423570985008687907853269, I think we are good there

philbow61 (approved):
LGTM nice work 🥇

PR #596, opened 2025-05-13
Feat: Liquidity Strategies

bayological:
### Description

This PR adds an extensible “liquidity strategy” framework to Mento v3 and a concrete Reserve-based implementation. The new contracts decouple the core rebalancing flow from specific liquidity sources, enabling future strategies (CDP, custom, cross-chain, etc.) to be added easily.

Key features:
- `ILiquidityStrategy` interface with a unified rebalance(pool) entry point and lifecycle events.
- `LiquidityStrategy` abstract base handling pool registration, cooldowns, price/threshold checks, and skip logic.
- `ReserveLiquidityStrategy` implementation that uses the on-chain Reserve:
  - Performs flash-swap rebalances (contraction/expansion) by minting/burning Mento stables and moving collateral.
  - Emits `RebalanceInitiated` and `RebalanceExecuted` for full traceability.

### Tested
- Unit tests covering the abstract `LiquidityStrategy` contract
- Unit tests covering the `ReserveLiquidityStrategy` - WIP

### DIFF
--- a/contracts/interfaces/ILiquidityStrategy.sol
+++ b/contracts/interfaces/ILiquidityStrategy.sol
@@ -18,8 +18,9 @@ interface ILiquidityStrategy {
    */
   struct FPMMConfig {
     uint256 lastRebalance;
-    uint256 rebalanceCooldown;
-    uint256 rebalanceThreshold; // TODO: Do we need more granularity here? (upper & lower).
+    uint256 rebalanceCooldown; // seconds
+    // TODO: Do we need more granularity here? (upper & lower).
+    uint256 rebalanceThreshold; // 18-decimal fixed point (e.g., 0.05e18 = 5%)
   }
 
   /**
@@ -42,7 +43,7 @@ interface ILiquidityStrategy {
    * @param priceBefore The pool price before the rebalance.
    * @param priceAfter The pool price after the rebalance.
    */
-  event Rebalance(address indexed pool, uint256 priceBefore, uint256 priceAfter);
+  event RebalanceExecuted(address indexed pool, uint256 priceBefore, uint256 priceAfter);
 
   /**
    * @notice Emitted when a rebalance is skipped because the cooldown period has not elapsed.
@@ -55,12 +56,6 @@ interface ILiquidityStrategy {
    * @param pool The address of the pool that was skipped.
    */
   event RebalanceSkippedPriceInRange(address indexed pool);
-
-  /**
-   * @notice Emitted when the reserve address is set.
-   * @param reserve The address of the reserve that was set.
-   */
-  event ReserveSet(address indexed reserve);
 
   /**
    * @notice Triggers the liquidity rebalancing mechanism.
--- a/contracts/interfaces/ILiquidityStrategy.sol
+++ b/contracts/interfaces/ILiquidityStrategy.sol
@@ -3,20 +3,64 @@ pragma solidity ^0.8.0;
 
 interface ILiquidityStrategy {
   /**
-   * @notice Struct holding the state of a pool.
-   * @param lastRebalance The timestamp of the last rebalance for this pool.
+   * @notice Enum representing the direction of price deviation from the oracle price.
+   * @dev Used to determine the direction of the price movement when rebalancing.
    */
-  struct PoolState {
+  enum PriceDirection {
+    ABOVE_ORACLE,
+    BELOW_ORACLE
+  }
+  /**
+   * @notice Struct holding the configuration of an FPMM pool.
+   * @param lastRebalance The timestamp of the last rebalance for this pool.
+   * @param rebalanceCooldown The cooldown period for the next rebalance.
+   * @param rebalanceThreshold The threshold of the price change that triggers a rebalance.
+   */
+  struct FPMMConfig {
     uint256 lastRebalance;
+    uint256 rebalanceCooldown;
+    uint256 rebalanceThreshold; // TODO: Do we need more granularity here? (upper & lower).
   }
 
   /**
-   * @notice Emitted when a pool is rebalanced.
+   * @notice Emitted after an FPMM pool is added.
+   * @param pool The address of the pool that was added.
+   * @param rebalanceThreshold The threshold of the price change that triggers a rebalance.
+   * @param rebalanceCooldown The cooldown period for the next rebalance.
+   */
+  event FPMMPoolAdded(address indexed pool, uint256 rebalanceThreshold, uint256 rebalanceCooldown);
+
+  /**
+   * @notice Emitted when an FPMM pool is removed.
+   * @param pool The address of the pool that was removed.
+   */
+  event FPMMPoolRemoved(address indexed pool);
+
+  /**
+   * @notice Emitted when an FPMM pool is rebalanced.
    * @param pool The address of the pool that was rebalanced.
    * @param priceBefore The pool price before the rebalance.
    * @param priceAfter The pool price after the rebalance.
    */
   event Rebalance(address indexed pool, uint256 priceBefore, uint256 priceAfter);
+
+  /**
+   * @notice Emitted when a rebalance is skipped because the cooldown period has not elapsed.
+   * @param pool The address of the pool that was skipped.
+   */
+  event RebalanceSkippedNotCool(address indexed pool);
+
+  /**
+   * @notice Emitted when a rebalance is skipped because the price is within the threshold.
+   * @param pool The address of the pool that was skipped.
+   */
+  event RebalanceSkippedPriceInRange(address indexed pool);
+
+  /**
+   * @notice Emitted when the reserve address is set.
+   * @param reserve The address of the reserve that was set.
+   */
+  event ReserveSet(address indexed reserve);
 
   /**
    * @notice Triggers the liquidity rebalancing mechanism.
--- a/contracts/interfaces/ILiquidityStrategy.sol
+++ b/contracts/interfaces/ILiquidityStrategy.sol
@@ -1,0 +1,26 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+pragma solidity ^0.8.0;
+
+interface ILiquidityStrategy {
+  /**
+   * @notice Struct holding the state of a pool.
+   * @param lastRebalance The timestamp of the last rebalance for this pool.
+   */
+  struct PoolState {
+    uint256 lastRebalance;
+  }
+
+  /**
+   * @notice Emitted when a pool is rebalanced.
+   * @param pool The address of the pool that was rebalanced.
+   * @param priceBefore The pool price before the rebalance.
+   * @param priceAfter The pool price after the rebalance.
+   */
+  event Rebalance(address indexed pool, uint256 priceBefore, uint256 priceAfter);
+
+  /**
+   * @notice Triggers the liquidity rebalancing mechanism.
+   * @param pool The address of the pool to rebalance.
+   */
+  function rebalance(address pool) external;
+}
--- a/contracts/interfaces/ILiquidityStrategy.sol
+++ b/contracts/interfaces/ILiquidityStrategy.sol
@@ -38,6 +38,22 @@ interface ILiquidityStrategy {
   event FPMMPoolRemoved(address indexed pool);
 
   /**
+   * @notice Emitted when a rebalance is initiated by the strategy.
+   * @param pool The pool being rebalanced.
+   * @param stableOut Amount of stable token being taken out of the pool.
+   * @param collateralOut Amount of collateral being taken out of the pool.
+   * @param inputAmount The amount the strategy is supplying in the callback.
+   * @param direction ABOVE_ORACLE for contraction, BELOW_ORACLE for expansion.
+   */
+  event RebalanceInitiated(
+    address indexed pool,
+    uint256 stableOut,
+    uint256 collateralOut,
+    uint256 inputAmount,
+    PriceDirection direction
+  );
+
+  /**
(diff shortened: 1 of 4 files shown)


contracts/swap/ReserveLiquidityStrategy.sol
```diff
@@ -0,0 +1,136 @@
...
+    fpm.rebalance(stableOut, collateralOut, address(this), callbackData);
+  }
+
+  /**
+   * @notice Handles the callback from the pool after a rebalance.
+   * @param amount0Out The amount of token0 to move out of the pool.
+   * @param amount1Out The amount of token1 to move out of the pool.
+   * @param data The encoded data from the pool.
+   */
+  function hook(address, uint256 amount0Out, uint256 amount1Out, bytes calldata data) external {
```
baroooo, line 109: I think there is an interesting vulnerability here. we currently allow a potential exploit where an attacker can call the fpmm.swap from any account with a callback that will burn/mint tokens on the reserve LS

contracts/swap/LiquidityStrategy.sol
```diff
@@ -0,0 +1,131 @@
...
+  function rebalance(address pool) external nonReentrant {
+    require(isPoolRegistered(pool), "LS: UNREGISTERED_POOL");
+
+    FPMMConfig memory config = fpmmPoolConfigs[pool];
+    if (config.lastRebalance > 0 && block.timestamp <= config.lastRebalance + config.rebalanceCooldown) {
+      emit RebalanceSkippedNotCool(pool);
+      return;
+    }
+
+    IFPMM fpm = IFPMM(pool);
```
baroooo, line 77: fpm -> fpmm?

baroooo (changes requested):

contracts/swap/LiquidityStrategy.sol
```diff
@@ -0,0 +1,131 @@
...
+    (uint256 oraclePrice, uint256 poolPrice) = fpm.getPrices();
```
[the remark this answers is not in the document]
baroooo: main reason was to stay consistent/compatible with aerodrome/uniswap pools

contracts/interfaces/ILiquidityStrategy.sol
```diff
@@ -0,0 +1,99 @@
...
+
+  /**
+   * @notice Emitted when a rebalance is initiated by the strategy.
+   * @param pool The pool being rebalanced.
+   * @param stableOut Amount of stable token being taken out of the pool.
+   * @param collateralOut Amount of collateral being taken out of the pool.
+   * @param inputAmount The amount the strategy is supplying in the callback.
+   * @param direction ABOVE_ORACLE for contraction, BELOW_ORACLE for expansion.
+   */
+  event RebalanceInitiated(
```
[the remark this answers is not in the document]
bayological: The separation helps track the lifecycle of the rebalance operation. There are multiple points of failure here so this makes tracking easier. `RebalanceInitiated` is to signal the start of the rebalance and `RebalanceExecuted` confirms the completion and reports the outcome.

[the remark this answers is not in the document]
philbow61: I don't think we need the initiation event. If the rebalancing doesn't go through, the initiation event will be rolled back anyway. The FPMM is also emitting one event on rebalancing. I would suggest let's remove the RebalanceInitiated one and change the RebalanceExecuted to emit info we are not also getting from the FPMM, e.g stables in/out, collateral in/out.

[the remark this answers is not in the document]
bayological: I agree with possibly improving the info logged with RebalancedExecuted, but I'm in favour of keeping the event. 

Think about this scenario:
- Rebalance initiated is emitted 
- The call to Fpm.rebalance() is successful
- The fpm then calls the hook function
- The reserve.transferExchangeCollateralAsset() fails

Will the initiated event be rolled back?

contracts/swap/ReserveLiquidityStrategy.sol
```diff
@@ -0,0 +1,201 @@
...
+   */
+  function hook(address sender, uint256 amount0Out, uint256 amount1Out, bytes calldata data) external {
+    (address pool, uint256 amountIn, PriceDirection priceDirection) = abi.decode(
+      data,
+      (address, uint256, PriceDirection)
+    );
+
+    require(sender == address(this), "RLS: HOOK_SENDER_NOT_SELF");
+    require(msg.sender == pool, "RLS: HOOK_CALLER_NOT_EXPECTED_POOL");
+    require(isPoolRegistered(pool), "RLS: UNREGISTERED_POOL");
```
baroooo, line 73: I think isPoolRegistered(msg.sender) should be enough and we can remove pool from the callback if it does not serve another purpose

contracts/swap/ReserveLiquidityStrategy.sol
```diff
@@ -0,0 +1,201 @@
...
+
+    if (priceDirection == PriceDirection.ABOVE_ORACLE) {
+      (collateralOut, inputAmount) = _calculateExpansionAmounts(reserves, ud(oraclePrice));
+      stableOut = 0;
+    } else {
+      (stableOut, inputAmount) = _calculateContractionAmounts(reserves, ud(oraclePrice));
+      collateralOut = 0;
+    }
+
+    bytes memory callbackData = abi.encode(pool, inputAmount, priceDirection);
```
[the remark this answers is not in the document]
bayological: You are right, this was added before the Fpmm implementation was done. Will remove it.

contracts/swap/LiquidityStrategy.sol
```diff
@@ -0,0 +1,132 @@
...
+    (uint256 oraclePrice, uint256 poolPrice, , ) = fpmm.getPrices();
+    require(oraclePrice > 0 && poolPrice > 0, "LS: INVALID_PRICES");
+
+    uint256 rawUpperThresholdBps = fpmm.rebalanceThresholdAbove();
+    require(rawUpperThresholdBps > 0 && rawUpperThresholdBps <= 10_000, "LS: INVALID_UPPER_THRESHOLD");
+
+    uint256 rawLowerThresholdBps = fpmm.rebalanceThresholdBelow();
+    require(rawLowerThresholdBps > 0 && rawLowerThresholdBps <= 10_000, "LS: INVALID_LOWER_THRESHOLD");
+
+    // Convert basis points to scaled decimals (e.g. 10000 bps = 1.0)
```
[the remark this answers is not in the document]
bayological: What would the result be if we used raw BPS values (e.g., 500) directly in the calculations?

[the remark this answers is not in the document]
baroooo: as long as we do multiplication first it should be ok

test/unit/swap/ReserveLiquidityStrategy.t.sol
```diff
@@ -0,0 +1,597 @@
...
+  function test_rebalance_whenCooldownActive_shouldRevert() public {
+    // Initial rebalance to set timestamp
+    // Create a scenario that will trigger expansion (price above oracle)
+    mockPool.setPrices(1e18, 2e18); // poolPrice > Oracle + threshold
+    mockPool.setReserves(2e21, 5e21); // S=2000, C=5000, P_oracle=1
+
+    // Mint tokens to support the test
+    collateralToken.mint(address(mockPool), 50e18);
+    collateralToken.mint(address(mockReserve), 50e18);
+
```
philbow61, line 169: These different values don't fit together here 
setReserves Price : 2.5 
poolPrice 2 
actualBuckets coll: 1050, stable 1000 
I now this is not relevant to the test case but still it's a bit confusing.

contracts/swap/ReserveLiquidityStrategy.sol
```diff
@@ -0,0 +1,195 @@
...
+    // Y = (C - P * S) / 2
+    // X = Y / P
+    uint256 numerator = reserves.collateralReserve - ((oraclePrice * reserves.stableReserve) / 1e18);
+    uint256 denominator = 2;
+
+    uint256 collateralOutRaw = numerator / denominator;
+    uint256 stablesInRaw = (numerator * 1e18) / (denominator * oraclePrice);
+
+    collateralOut = collateralOutRaw / reserves.collateralPrecision;
+    stablesIn = stablesInRaw / reserves.stablePrecision;
```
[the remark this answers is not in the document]
baroooo: I agree, a configurable incentive on the strategy contract that is bounded by the value set in the FPMM makes sense

test/unit/swap/ReserveLiquidityStrategy.t.sol
bayological: The goal of this test is just to verify that once a rebalance is executed a second rebalance can't happen. To test this we just need anything that will trigger the first rebalance. So even though these values don't reflect a real world scenario, for this case they trigger the rebalance, which is all we want. The exact reserve amounts etc are not what is being tested here.

contracts/swap/ReserveLiquidityStrategy.sol
```diff
@@ -0,0 +1,216 @@
...
+    IFPMM fpmm,
+    uint256 oraclePrice,
+    PriceDirection priceDirection
+  ) private view returns (uint256 stableOut, uint256 collateralOut, uint256 inputAmount) {
+    (uint256 dec0, uint256 dec1, uint256 reserve0, uint256 reserve1, , ) = fpmm.metadata();
+
+    require(dec0 <= 18 && dec1 <= 18, "RLS: TOKEN_DECIMALS_TOO_LARGE");
+
+    uint256 fpmmIncentive = fpmm.rebalanceIncentive();
+    uint256 incentive = rebalanceIncentive < fpmmIncentive ? rebalanceIncentive : fpmmIncentive;
```
philbow61, line 143: If we put the check in the setter, we can remove this line

contracts/swap/LiquidityStrategy.sol
```diff
@@ -44,13 +42,14 @@ abstract contract LiquidityStrategy is ILiquidityStrategy, OwnableUpgradeable, R
   /* ==================== Admin Functions ==================== */
 
   /// @inheritdoc ILiquidityStrategy
-  function addPool(address poolAddress, uint256 cooldown) external onlyOwner {
+  function addPool(address poolAddress, uint256 cooldown, uint256 rebalanceIncentive) external onlyOwner {
     require(poolAddress != address(0), "LS: INVALID_POOL_ADDRESS");
     require(fpmmPools.add(poolAddress), "LS: POOL_ALREADY_ADDED");
+    require(rebalanceIncentive > 0 && rebalanceIncentive <= BPS_SCALE, "LS: INVALID_REBALANCE_INCENTIVE");
```
philbow61, line 48: Two points here:
- The rebalance incentive can be 0
- instead of checking <= BPS_scale it makes sense to check <=  fpmm.rebalanceIncentive();
I know we are ensuring that also in the rebalance, but if someone tries setting a value larger than the max in the pool, we should still revert here.  To indicate the configuration is invalid.

contracts/swap/LiquidityStrategy.sol
```diff
@@ -0,0 +1,143 @@
...
+    delete fpmmPoolConfigs[pool];
+    emit FPMMPoolRemoved(pool);
+  }
+
+  /**
+   * @notice Sets the rebalance incentive in basis points.
+   * @param pool The address of the pool to set the rebalance incentive for.
+   * @param rebalanceIncentive The rebalance incentive in basis points.
+   */
+  function setRebalanceIncentive(address pool, uint256 rebalanceIncentive) public onlyOwner {
```
philbow61, line 71: We should also perform checks on the incentive in this function. Also, there is no check here whether the pool exists.

contracts/swap/ReserveLiquidityStrategy.sol
```diff
@@ -0,0 +1,221 @@
...
+    (
+      uint256 stableOut,
+      uint256 collateralOut,
+      uint256 inputAmount,
+      uint256 incentiveAmount
+    ) = _calculateRebalanceAmounts(fpmm, oraclePrice, priceDirection);
+
+    bytes memory callbackData = abi.encode(inputAmount, priceDirection, incentiveAmount);
+
+    emit RebalanceInitiated(pool, stableOut, collateralOut, inputAmount, priceDirection);
```
philbow61, line 133: I think it could be a bit confusing if we emit this event with the inputAmount param since the actual amount that goes into the fpmm will be inputAmount-incentiveAmount. Maybe we could emit inputAmount-incentiveAmount and incentiveAmount instead to show the full picture here

contracts/swap/LiquidityStrategy.sol
```diff
@@ -63,6 +65,21 @@ abstract contract LiquidityStrategy is ILiquidityStrategy, OwnableUpgradeable, R
...
+   * @dev This function can be used to retrieve collected fees.
+   * @param tokenAddress The address of the ERC20 token to withdraw.
+   */
+  function withdraw(address tokenAddress) external onlyOwner {
+    require(tokenAddress != address(0), "LS: INVALID_TOKEN_ADDRESS");
+    IERC20Upgradeable token = IERC20Upgradeable(tokenAddress);
+    uint256 balance = token.balanceOf(address(this));
+    require(balance > 0, "LS: NO_TOKENS_TO_WITHDRAW");
+
+    bool success = token.transfer(owner(), balance);
```
philbow61, line 79: I think we said we wan't to specify the to address in the function call.
Also if we move SafeErc20 into this contract we can use it in both this and the ReserveLiquiditystartegy and not do the transfer check manually

bayological: okk

PR #597, opened 2025-05-14
Feat/fpmm-factory-contract

philbow61:
### Description

This PR adds the factory contract that will deploy 1 FPMM implementation contract per chain as well as multiple FPMM proxies per pair. 
The contract uses the [CREATX](https://github.com/pcaversaccio/createx) contract, which is deployed to most networks.
The contract ensures same deployment addresses across different networks as long as this factory contract is deployed to the same address across networks.

### Other changes

- I pulled Barans latest changes since the factory needs the FPMM 😁, but for reviewing this P,R only the factory-related files are important 

### Tested

- unit tests
- fork tests to ensure fpmms are deployed to the same addresses


contracts/swap/FPMMFactory.sol
```diff
@@ -0,0 +1,246 @@
...
+      hash := keccak256(0x00, 0x40)
+    }
+  }
+
+  /**
+   * @dev This empty reserved space is put in place to allow future versions to add new
+   * variables without shifting down storage in the inheritance chain.
+   * See https://docs.openzeppelin.com/contracts/4.x/upgradeable#storage_gaps
+   */
+  uint256[50] private __gap;
```
baroooo, line 245: this should be 44, but lets discuss the pattern bogdan suggested first

contracts/swap/FPMMFactory.sol
```diff
@@ -0,0 +1,246 @@
...
+  }
+
+  // slither-disable-start encode-packed-collision
+  /// @inheritdoc IFPMMFactory
+  function getOrPrecomputeProxyAddress(address token0, address token1) public view returns (address) {
+    if (address(deployedFPMMs[token0][token1]) != address(0)) {
+      return address(deployedFPMMs[token0][token1]);
+    }
+    bytes11 customSalt = bytes11(
+      uint88(uint256(keccak256(abi.encodePacked(IERC20(token0).symbol(), IERC20(token1).symbol()))))
```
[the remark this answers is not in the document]
philbow61: yes good point 👍

contracts/swap/FPMMFactory.sol
```diff
@@ -0,0 +1,246 @@
...
+    bytes32 guardedSalt = _efficientHash({ a: bytes32(uint256(uint160(address(this)))), b: salt });
+
+    bytes memory initData = abi.encodeWithSelector(
+      FPMM.initialize.selector,
+      token0,
+      token1,
+      sortedOracles,
+      breakerBox,
+      governance
+    );
```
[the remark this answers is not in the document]
philbow61: yes ser 👍

contracts/swap/FPMMFactory.sol
```diff
@@ -0,0 +1,246 @@
...
+
+  /// @inheritdoc IFPMMFactory
+  function initialize(
+    address _sortedOracles,
+    address _proxyAdmin,
+    address _breakerBox,
+    address _governance
+  ) external initializer {
+    __Ownable_init();
+    setProxyAdmin(_proxyAdmin);
```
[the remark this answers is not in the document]
philbow61: The way I thought about chain expansion is that for each new chain, we will have one ProxyAdmin per chain configured for all governance-owned Proxies. Therefore, this ProxyAdmin will most likely already be deployed once we deploy the FPMMFactory for this chain's OracleProxy, BreakerBoxProxy ...

test/unit/swap/FPMM/FPMMFactory.t.sol
```diff
@@ -0,0 +1,500 @@
...
+
+    address expectedFPMMImplementation = factoryCelo.getOrPrecomputeImplementationAddress();
+    vm.expectEmit();
+    emit FPMMImplementationDeployed(expectedFPMMImplementation);
+    factoryCelo.deployFPMM(token0Celo, token1Celo);
+
+    assertEq(factoryCelo.fpmmImplementation(), expectedFPMMImplementation);
+
+    vm.expectRevert("Initializable: contract is already initialized");
+    FPMM(expectedFPMMImplementation).initialize(
```
baroooo, line 355: not related to test

test/unit/swap/FPMM/FPMMFactory.t.sol
```diff
@@ -0,0 +1,500 @@
...
+
+    vm.prank(deployer);
+    factoryOp = new FPMMFactory(false);
+    factoryOp.initialize(sortedOraclesOp, proxyAdminOp, breakerBoxOp, governanceOp);
+    deployCodeTo("ERC20", abi.encode("Token 0", "T0"), token0Op);
+    deployCodeTo("ERC20", abi.encode("Token 1", "T1"), token1Op);
+  }
+
+  function test_deployFPMM_shouldDeploySameFPMMToSameAddressOnDifferentChains() public {
+    // factoryCelo and factoryOp need to be the same address becuase of CREATEX permissioned deploy protection
```
baroooo, line 477: is this comment correct?

test/unit/swap/FPMM/FPMMFactory.t.sol
```diff
@@ -0,0 +1,500 @@
...
+    assertEq(address(factoryCelo), address(factoryOp));
+    vm.selectFork(celoFork);
+    vm.assertEq(vm.activeFork(), celoFork);
```
baroooo, line 480: this feels like testing forge itself :D

contracts/swap/FPMMFactory.sol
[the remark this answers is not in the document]
philbow61: Decision with the management: no use case we need to support

contracts/swap/FPMMFactory.sol
```diff
@@ -0,0 +1,267 @@
...
+   * @notice Deploys the FPMM proxy contract.
+   * @param token0 The address of the first token
+   * @param token1 The address of the second token
+   * @param _referenceRateFeedID The address of the reference rate feed
+   * @return The address of the deployed FPMM proxy
+   */
+  // slither-disable-start reentrancy-benign
+  // slither-disable-start reentrancy-events
+  // slither-disable-start encode-packed-collision
+  function _deployFPMMProxy(address token0, address token1, address _referenceRateFeedID) internal returns (address) {
```
chapati23, line 182: why the underscore prefix for referenceRateFeedID?

contracts/swap/FPMMFactory.sol
```diff
@@ -0,0 +1,276 @@
...
+
+  /// @inheritdoc IFPMMFactory
+  function deployFPMM(
+    address token0,
+    address token1,
+    address referenceRateFeedID
+  ) external onlyOwner returns (address, address) {
+    require(token0 != address(0) && token1 != address(0), "FPMMFactory: ZERO_ADDRESS");
+    require(token0 != token1, "FPMMFactory: IDENTICAL_TOKEN_ADDRESSES");
+    require(referenceRateFeedID != address(0), "FPMMFactory: ZERO_ADDRESS");
```
chapati23, line 151: Should we add a check for contract existence: `require(token0.code.length > 0, "FPMMFactory: NOT_A_CONTRACT");`

philbow61: I don't think this is really necessary because the _computeProxyAddressAndSalt() that is called as part of this function will call symbol() on these addresses, ensuring there is a contract that, at minimum supports symbol() is deployed to these addresses. So this call won't go through if the addresses are not a contract and don't support symbol(). 
The same is also enforced through the initialize function of the fppm that is called as a result of deployFPMM and requires them to support decimals() and symbol(). 
So, unless we want to have a clearer error message, I would prefer not to add it and have fewer require checks here, but let me know if you think it's important.

contracts/swap/FPMMFactory.sol
```diff
@@ -0,0 +1,276 @@
...
+
```
chapati23, line 152: Should we add a check that we don't try to redeploy an existing pair?

```sol
require(address(deployedFPMMs[token0][token1]) == address(0), "FPMMFactory: PAIR_ALREADY_EXISTS");
```

contracts/swap/FPMMFactory.sol
```diff
@@ -0,0 +1,276 @@
...
+   * @notice Computes the address of the FPMM proxy contract.
+   * @dev apply permissioned deploy protection with factory address and custom salt
+   *      see https://github.com/pcaversaccio/createx?tab=readme-ov-file for more details
+   *      custom salt is a keccak256 hash of the token0 and token1 symbols
+   * @param token0 The address of the first token
+   * @param token1 The address of the second token
+   * @return The precomputed address of the FPMM proxy contract
+   * @return The salt used to deploy the FPMM proxy contract
+   */
+  function _computeProxyAddressAndSalt(address token0, address token1) internal view returns (address, bytes32) {
```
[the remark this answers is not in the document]
chapati23: Yeah, I was mainly thinking about a potential migration from one rate feed to another. If the only path we'll need is `FPMM.setRateFeedId()`, then we're good w/o the rate feed ID in the salt.

contracts/swap/FPMMFactory.sol
```diff
@@ -0,0 +1,383 @@
...
+    $.isRegisteredImplementation[fpmmImplementation] = false;
+    if ($.registeredImplementations.length > 1) {
+      $.registeredImplementations[index] = $.registeredImplementations[$.registeredImplementations.length - 1];
+    }
+    $.registeredImplementations.pop();
+    emit FPMMImplementationUnregistered(fpmmImplementation);
+  }
+
+  /// @inheritdoc IFPMMFactory
+  function deployFPMM(
```
bayological, line 212: Two thoughts on this regarding the tokens. 
1. We said that token0 would always be the stable token and token0 the collateral. However, we don't enforce this anywhere. Unless there's reason not to, I think we should be doing this here. Either that or we trust that people deploying will always adhere to this convention.

2. Kind of related to the first point but, right now we can do this:
`deployFPMM(USDC, cUSD....)`
then create another like this
`deployFPMM(cUSD, USDC....)` ?
These should be the same pool but would end up as two separate ones. If we do enforce the convention then this would not happen, but is there a case for creating a pool where a token isn't a Mento stable? (e.g. USDC <-> USDT)

bayological (approved):

PR #604, opened 2025-06-27
fix: alfajores fork tests post pool restructuring

nvtaveras:
### Description

This fixes the alfajores fork tests after executing the pool restructuring CGP. We now have 19 total exchanges, and a new dependency on the XOF/USD feed.


bayological (approved):

philbow61 (commented):
Looks great

PR #609, opened 2025-08-21
refactor: Improved liquidity management

bayological:
### Description

The previous implementation of the liquidity strategies followed a monolithic design, that had many flaws, but also made it harder to implement additional strategies. 

The intention of this refactor is to address the flaws and to create a more robust liquidity management system. 

This change also introduces the concept of a 'Liquidity Policy Pipeline'. In the previous implementation, and the spec, a pool could only have one liquidity strategy for it's rebalancing. 

In this new implementation, it is now also possible for a pool to have multiple strategies configured that will execute their rebalancing operations sequentially. It's worth noting that it's not **required** for a pool to have multiple strategies but it is now supported in both the fpmm and the liquidity system.

The new implementation follows a more modular design with the following main components:
- **Liquidity Controller:** The central 'orchestrator' managing pool configurations, policy pipelines and liquidity strategy executions
- **ILiquidityPolicy**: The policy is a pluggable decision making interface. It should determine if a rebalance should happen, and if so come up with the amounts that should be moved.  A policy must have a linked strategy
- **ILiquidityStrategy**: The liquidity strategy is the execution engine for it's linked policy. It interfaces with the fpmm to execute the movement of tokens based on the action described by the policy

### Other changes

None

### Tested

Added unit tests for the new contracts.


contracts/v3/LiquidityController.sol
```diff
@@ -0,0 +1,293 @@
...
+
+      require(oracleNum > 0 && oracleDen > 0, "LC: INVALID_PRICES");
+
+      ctx.reserves = LQ.Reserves({ reserveNum: reserveNum, reserveDen: reserveDen });
+      ctx.prices = LQ.Prices({
+        oracleNum: oracleNum,
+        oracleDen: oracleDen,
+        poolPriceAbove: poolAbove,
+        diffBps: diffBps
+      });
```
bowd, line 251: This is a bit confusing to me, IIUC reserveNum and reserveDen is the pool's implicit price (i.e. ratio of reserve assets) but this makes it sound like a reserve amount. I would suggest changing the naming to:
- implicitPriceNum and implicitPriceDen
- oraclePriceNum and oraclePriceDen
And I'm not sure why both shouldn't live in the Prices struct.

contracts/v3/ReservePolicy.sol
```diff
@@ -0,0 +1,171 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
```
bowd, line 1: I'm unsure what the difference would be between the ReservePolicty and the CDPPolicy, reading through it it feels like the calculations would be the same here.

contracts/v3/LiquidityController.sol
```diff
@@ -0,0 +1,293 @@
...
+  function setPoolPipeline(address pool, ILiquidityPolicy[] calldata policies) external onlyOwner {
+    _ensurePool(pool);
+    delete pipelines[pool];
+    address[] memory policyAddresses = new address[](policies.length);
+    for (uint256 i = 0; i < policies.length; i++) {
+      pipelines[pool].push(policies[i]);
+      policyAddresses[i] = address(policies[i]);
+    }
+    emit PipelineSet(pool, policyAddresses);
+  }
```
bowd, line 124: I'm wondering a bit why we need multiple policies pre pool, given my comment on the ReservePolicy.sol where it feels like that policy would work for CDP rebalancing as well. I'm leaning towards removing this, also removing the mapping between liquidity source => liquidity strategy. Just have the ReservePoliciy as a library or part of this contract, and have pool => liquidity strategy mapping for executing the action which can be part of the config. wdyt?

contracts/v3/LiquidityController.sol
```diff
@@ -0,0 +1,293 @@
...
+    for (uint256 i = 0; i < policies.length; i++) {
+      (bool shouldAct, LQ.Action memory action) = policies[i].determineAction(ctx);
+      if (!shouldAct) continue;
+
+      ILiquidityStrategy strategy = strategies[action.liquiditySource];
+      require(address(strategy) != address(0), "LC: NO_STRATEGY_FOR_LIQUIDITY_SOURCE");
+
+      bool ok = strategy.execute(action);
+      require(ok, "LC: STRATEGY_EXECUTION_FAILED");
+      acted = true;
```
bowd, line 161: nit but we could just do `acted = strategy.execute(action)` here and `require(acted, ...)`

contracts/v3/LiquidityController.sol
```diff
@@ -0,0 +1,293 @@
...
+
+  /// @inheritdoc ILiquidityController
+  function addPool(
+    address pool,
+    address debtToken,
+    address collateralToken,
+    uint64 cooldown,
+    uint32 incentiveBps
+  ) external onlyOwner {
+    require(pool != address(0), "LC: POOL_MUST_BE_SET");
```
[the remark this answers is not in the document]
bayological: Custom errors are great but I stuck with string reverts to be consistent with the V3 contracts(FPMM, FPMMFactory etc). So whichever we use, I agree we should be consistent.

contracts/v3/LiquidityController.sol
bayological: I'm against this. You're basically suggesting we go back to the initial implementation. 

We currently don't have a need for multiple polices per pool, but this design was intentional to give us that option in the future. Maybe if execution of the reserve strategy does not bring the price back in range we could fall back to the cdp strategy?.  The Circuit Breaker also has this level of modularity, which wasn't required initially, but I vaguely remember us being able to use the flexibility to have multiple median delta breakers for something.

On the ReservePolicy + CDP Policy point, when we started this work, there were actually two different formulas for CDP and reserve rebalancing. It was only recently that P1 came up with a new formula that could be used for both. But while they've converged for now, they could diverge again as things evolve. This current design has a clean separation, which would allow us to change things easily and in isolation (e.g. kind of like how we were able to add the market hours breaker)

In general though, I'd prefer to keep the design as is. There are a few mappings and a slight gas overhead but I think the benefits we get in testability and future extensibility are worth it.

contracts/v3/LiquidityController.sol
```diff
@@ -0,0 +1,293 @@
...
+    require(debtToken != address(0) && collateralToken != address(0), "LC: TOKENS_MUST_BE_SET");
+    require(debtToken != collateralToken, "LC: TOKENS_MUST_BE_DIFFERENT");
+
+    // Verify LC token ordering matches FPMM ordering
+    (address expectedToken0, address expectedToken1) = _orderTokens(debtToken, collateralToken);
+    (, , , , address actualT0, address actualT1) = IFPMM(pool).metadata();
```
Mouradif, line 67: there's also `IFPMM.tokens()` if we only care about the 2 tokens addresses in the correct order

contracts/v3/LiquidityController.sol
```diff
@@ -0,0 +1,293 @@
...
+      bool ok = strategy.execute(action);
+      require(ok, "LC: STRATEGY_EXECUTION_FAILED");
+      acted = true;
+
+      // refresh after action execution, stop early if price in range
+      (ctx, inRange, ) = _readCtx(pool, config);
+      if (inRange) break;
+    }
+
+    if (acted) {
```
Mouradif, line 168: I vote we revert if `!acted`

contracts/v3/ReserveLiquidityStrategy.sol
```diff
@@ -0,0 +1,260 @@
...
+  function _handleRebalanceCallback(
+    address pool,
+    uint256 amount0Out,
+    uint256 amount1Out,
+    uint256 inputAmount,
+    uint256 incentiveAmount,
+    LQ.Direction direction,
+    bool isToken0Debt
+  ) internal {
+    (, , , , address token0, address token1) = IFPMM(pool).metadata();
```
Mouradif, line 178: same as previous comment: `IFPMM(pool).tokens()`. Here, since the caller is the pool itself, I wonder if it couldn't just pass those tokens as calldata and save us a 100 gas call

bowd (approved):

PR #612, opened 2025-08-29
feat: Trading limits V2

bayological:
### Description

This PR introduces a new trading limits library, TradingLimitsv2. 

The implementation is similar to V1 with the following key changes:
- Removed LG
- t0 and t1 are now fixed (5mins & 1day respectively)
- Netflows are now stored with 15 decimals
- Uses int96 for netflows and int112 for limits (vs int48 in V1) to support higher precision and larger values

### Other changes

- Integrated TradingLimitsV2 with FPMM

### Tested

- Added unit tests


contracts/libraries/TradingLimitsV2.sol
```diff
@@ -0,0 +1,215 @@
...
+  ) internal view returns (ITradingLimitsV2.State memory) {
+    if (_deltaFlow == 0) {
+      return self;
+    }
+
+    int96 scaledDelta = scaleFlow(_deltaFlow, decimals);
+
+    if (scaledDelta == 0) {
+      return self;
+    }
```
bowd, line 132: I'd say we want to round up or down to 1e3 (if _deltaFlow >0) or -1e3 (if _deltaFlow < 0) in this case.

contracts/libraries/TradingLimitsV2.sol
```diff
@@ -0,0 +1,215 @@
...
+  }
+
+  /**
+   * @notice Scale a limit from token decimals to internal precision (15 decimals).
+   * @dev Used to scale config limits for comparison with state netflows.
+   * @param limit the limit amount in token decimals.
+   * @param decimals the token's decimal places.
+   * @return the scaled limit in 15 decimal precision.
+   */
+  function scaleLimit(int112 limit, uint8 decimals) internal pure returns (int96) {
```
bowd, line 184: I'm not sure we need both, we can just have a `scaleValue(int256 value, uint8 decimals) internal pure returns (int96)` and call it with `scaleValue(_delatFlow, decimals)` and `scaleValue(int256(limit), decimals)`

contracts/swap/FPMM.sol
```diff
@@ -726,4 +771,38 @@ contract FPMM is IFPMM, ReentrancyGuardUpgradeable, ERC20Upgradeable, OwnableUpg
...
+  function _applyTradingLimits0(uint256 amountIn, uint256 amountOut) internal {
+    FPMMStorage storage $ = _getFPMMStorage();
+
+    if ($.tradingLimitsConfig0.flags > 0) {
+      int256 deltaFlow = int256(amountOut) - int256(amountIn);
+      uint8 decimals = ERC20Upgradeable($.token0).decimals();
+
+      $.tradingLimitsState0 = $.tradingLimitsState0.update($.tradingLimitsConfig0, deltaFlow, decimals);
+      $.tradingLimitsState0.verify($.tradingLimitsConfig0, decimals);
+    }
```
bowd, line 789: You can optimise things here a bit by loading only once in memory and then form memory to storage at the end:
```solidity
ITradingLimitsV2.State memory state = $.tradingLimitsState0;
ITradingLimitsV2.Config memory config = $.tradingLimitsConfig0;

if (config.flags > 0) {
  // ...
  state = state.update(config, deltaFlow, decimals);
  state.verify(config, decimals); // will revert if exceeded
  $.tradingLimitsState0 = state; // write at the end
}
```

Also very nitty, but if we massage the data layout a bit in ITradingLimitsV2.sol:

```solidity
struct TradingLimits {
  State state;
  Config config;
}
```

Then in the FPMM we can have:
```solidity
mapping(address token => ITradingLimitsV2.TradingLimits) tradingLimits;
```
^ have a single mapping instead of the 4 storage slots. 

We can have a single `_applyTradingLimits(address token, uint256 amountIn, uint256 amountOut);`
```solidity
ITradingLimitsV2.State memory state = $.tradingLimits[token].state;
ITradingLimitsV2.Config memory config = $.tradingLimits[token].config;
// ...
$.tradingLimits[token].state = state; // write at the end
// ... 
```

And call `_applyTradingLimits(token0, amount0In, amount0out)` and ... token1, amount1In, amount1Out

bowd (changes requested):
Looks good overall, only one slightly more serious comment regarding rounding, and one regarding performance (read to memory -> write to storage at the end), the rest with the extra struct and mapping I think are nice but not a hill I would die on, but if you like it go for it.

contracts/swap/FPMM.sol
```diff
@@ -726,4 +762,16 @@ contract FPMM is IFPMM, ReentrancyGuardUpgradeable, ERC20Upgradeable, OwnableUpg
...
+  /**
+   * @notice Apply trading limits for a token
+   * @param token Address of the token
+   * @param amountIn Amount of token flowing into the pool
+   * @param amountOut Amount of token flowing out of the pool
+   */
+  function _applyTradingLimits(address token, uint256 amountIn, uint256 amountOut) internal {
+    FPMMStorage storage $ = _getFPMMStorage();
+    uint8 decimals = ERC20Upgradeable(token).decimals();
+    $.tradingLimits[token] = $.tradingLimits[token].applyTradingLimits(amountIn, amountOut, decimals);
```
bowd, line 775: While this looks nice, we'll be forcing two SSTOREs here for the config and the state, and we should avoid that as the config hasn't changed. Let's make `applyTradingLimits` return only the state, and set only that in storage.

Mouradif (commented):
Non-blocking comment but I suggest unless we have a good reason to prevent setting L1 without L0, to just get rid of that `require()`

bowd (approved):

contracts/libraries/TradingLimitsV2.sol
```diff
@@ -0,0 +1,218 @@
...
+  int96 private constant MAX_INT96 = type(int96).max;
+  int96 private constant MIN_INT96 = type(int96).min;
+
+  /**
+   * @notice Validate a trading limit configuration.
+   * @dev Reverts if the configuration is malformed.
+   * @param self the Config struct to check.
+   */
+  function validate(ITradingLimitsV2.Config memory self) internal pure {
+    require(self.flags & L1 == 0 || self.flags & L0 != 0, "L1 without L0 not allowed");
```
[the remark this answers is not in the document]
bayological: Sure, I'll defer to @bowd on this since this was his design, so not sure if there was some reason behind that.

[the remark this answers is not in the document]
bowd: I agree with @Mouradif's assesment here, that's exactly why the initial design was like this, when they're fully configurable it doesn't make sense to have L2 without L1, but in this case, I think it's fine tot have one without the other.

PR #615, opened 2025-09-03
feat: adaptore 🤝 fpmm integration

nvtaveras:
### Description

This adds the adaptore contract and integrates it into the FPMM. A few things about this PR:

1. Chainlink recommended market hours are 18:00 ET Sunday to 17:00 ET Friday [(source)](https://docs.chain.link/data-feeds/selecting-data-feeds#market-hours). We decided not to overcomplicate the implementation to consider daylight savings so the "closed" market hours are `[Friday 21:00 UTC, Sunday 23:00 UTC)` which safely covers both EST and EDT.
2. Holidays included were Christmas and NYE, also as specified by Chainlink
3. ~~I had to create a new `WithCoolDownV2` contract which is essentially the same as the original with a newer solidity version, to be able to also bump the sol version on the new breaker.~~ ended up removing this
4. We discussed changing the contracts to use the custom storage layout support on sol >= `0.8.29` [(source)](https://soliditylang.org/blog/2025/03/12/solidity-0.8.29-release-announcement/) but this is not yet implemented in order to keep consistent with what's currently on the repo at the time of this PR

note: ~~there shouldn't be any new slither warnings introduced from this PR and all of the existing warnings should be fixed once #610 is merged~~ now fixed

### Tested

Wrote unit tests for the new contracts and updated the FPMM tests to integrate the Adaptore


contracts/swap/FPMM.sol
```diff
@@ -374,10 +364,10 @@ contract FPMM is IFPMM, ReentrancyGuardUpgradeable, ERC20Upgradeable, OwnableUpg
     require(amount0Out > 0 || amount1Out > 0, "FPMM: INSUFFICIENT_OUTPUT_AMOUNT");
     require(amount0Out < $.reserve0 && amount1Out < $.reserve1, "FPMM: INSUFFICIENT_LIQUIDITY");
     require(to != $.token0 && to != $.token1, "FPMM: INVALID_TO_ADDRESS");
-    require(
-      $.breakerBox.getRateFeedTradingMode($.referenceRateFeedID) == TRADING_MODE_BIDIRECTIONAL,
-      "FPMM: TRADING_SUSPENDED"
-    );
+
```
philbow61, line 367: Instead of making these three external calls to the adaptore we could add a single function on the adaptore side that returns all three informations together with the rateNum and rateDen. 
This would result in 1 instead of 4 external calls and 1 instead of 4 storage reads on the adaptore side. Potentially, all of this could happen in _getRateFeed(). This would result in getPrices() to revert when anything is up with the rateFeed, which i think is fair.

nvtaveras: updated based on our in-person discussion

contracts/oracles/breakers/MarketHoursBreaker.sol
```diff
@@ -0,0 +1,62 @@
...
+
+  /// @inheritdoc IMarketHoursBreaker
+  function isMarketOpen(uint256 timestamp) public pure returns (bool) {
+    return !_isWeekendHours(timestamp) && !_isHoliday(timestamp);
+  }
+
+  /// @inheritdoc IMarketHoursBreaker
+  // solhint-disable-next-line no-unused-vars
+  function shouldTrigger(address rateFeedID) public view returns (bool triggerBreaker) {
+    require(isMarketOpen(block.timestamp), "MarketHoursBreaker: Market is closed");
```
bayological, line 25: Is the intention that we'd only ever have this breaker configured for an FX rate feed? As if that is not the case, this revert would prevent any other configured breakers for the rate feed from being checked

contracts/oracles/breakers/MarketHoursBreaker.sol
```diff
@@ -0,0 +1,62 @@
...
+  /* ========== VIEW FUNCTIONS ========== */
+
+  /// @inheritdoc IMarketHoursBreaker
+  function isMarketOpen(uint256 timestamp) public pure returns (bool) {
+    return !_isWeekendHours(timestamp) && !_isHoliday(timestamp);
+  }
+
+  /// @inheritdoc IMarketHoursBreaker
+  // solhint-disable-next-line no-unused-vars
+  function shouldTrigger(address rateFeedID) public view returns (bool triggerBreaker) {
```
bayological, line 24: This is nice as it stops reports when the market is closed.

bayological (approved):
LGTM. Nice work

Mouradif (approved):
ok LGTM, just one last thing I have to point out, it seems impossible to achieve abstraction over the concept of Adaptore without the specificity of market hours for FX, so I suggest we reflect that in the name of the contract. At the end of the day, this is a contract that holds:
- breakerbox (kind of an oracle)
- market hours (which is kind of a breaker)
- sorted oracles

and its functionality is specific to FX, so why not FXOracleAdapter? Or maybe something even more market-hours-specific.

contracts/oracles/OracleAdapter.sol
```diff
@@ -0,0 +1,178 @@
...
+  /// @inheritdoc IAdaptore
+  function hasRecentRate(address rateFeedID) external view returns (bool) {
+    return _hasRecentRate(rateFeedID);
+  }
+
+  /* ========== INTERNAL FUNCTIONS ========== */
+
+  function _isMarketOpen() private view returns (bool) {
+    AdaptoreStorage storage $ = _getStorage();
+
```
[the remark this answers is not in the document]
nvtaveras: you brought up a very good point here! I chatted with Bogdan about this and decided to have two different functions: `getRateIfValid` which checks for trading mode and staleness, and `getFXRateIfValid` which does the same check plus the `isFXMarketOpen()` check.

So the current FPMM implementation will use the `getFXRateIfValid` call and it might be the case that we have a different implementation later on that uses `getRateIfValid` instead.

PR #619, opened 2025-09-22
Gas savings on FPMMFactory deployment by using `extcodehash`

Mouradif:
### Description

To ensure we have the right `CreateX` contract, we perform a check on its codehash. Previously, this was done by loading the entire code at the target address into memory and hashing it, this update makes use of the `extcodehash` opcode which allows us to make huge gas savings (4357 gas units in this case).

### Tests & Backwards compatibility

Absolutely no side effect is expected.


Mouradif:
@baroooo @bowd Please approve again

I messed up with the previous PR, I merged without squasing so we had an extra ugly commit in the history of `feat/mento-v3`. This might not be the cleanest way but after trying to cleanly drop the extra commit and failing the letting ChatGPT mess my git and miserably failing too, I've gone ahead and force pushed `mento-v3` back to before that PR was merged, now we can merge this on top and I won't forget to squash 😅

PR #628, opened 2025-10-14
refactor: remove hardcoded params from FPMM

nvtaveras:
### Description

This updates the FPMM to remove the hardcoded initialization parameters (lp/protocol fee, rebalance incentives/thresholds) by adding a new struct which is passed during initialization:

```solidity
  struct FPMMParams {
    uint256 lpFee;
    uint256 protocolFee;
    address protocolFeeRecipient;
    uint256 rebalanceIncentive;
    uint256 rebalanceThresholdAbove;
    uint256 rebalanceThresholdBelow;
  }
  ``` 

This also updates the FPMMFactory to have a "default params" object which is used on standard deployments, while at the same time supporting custom params deployments.

### DIFF
--- a/.github/workflows/lint_test.yaml
+++ b/.github/workflows/lint_test.yaml
@@ -26,7 +26,9 @@ jobs:
           submodules: recursive
 
       - name: Install Foundry
-        uses: foundry-rs/foundry-toolchain@9979a41b5daa5da1572d973d7ac5a3dd2afc0221 # v1.3.5
+        uses: foundry-rs/foundry-toolchain@v1
+        with:
+          version: v1.3.5
 
       - name: Install Node.js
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: contracts/swap/FPMM.sol [hunk 1]
```diff
@@ -65,7 +65,8 @@ contract FPMM is IRPool, IFPMM, ReentrancyGuardUpgradeable, ERC20Upgradeable, Ow
     address _oracleAdapter,
     address _referenceRateFeedID,
     bool _invertRateFeed,
-    address owner_
+    address _initialOwner,
+    FPMMParams calldata _params
```
baroooo: how did we decide which params to go into the struct and which params to stay as individual params?

PR #637, opened 2025-10-22
feat: re-add trading limits

bayological:
### Description

Re-adds the trading limit V2 integration with the FPMM.

[Original PR](https://github.com/mento-protocol/mento-core/pull/612)
[Revert Commit](https://github.com/mento-protocol/mento-core/compare/361b5b1e95408164084900babc7da249f926f6aa..9725c5ef7498f77cd0b706a3b4d48e319155e511)

### DIFF
--- a/test/unit/swap/FPMM/FPMMAdmin.t.sol
+++ b/test/unit/swap/FPMM/FPMMAdmin.t.sol
@@ -301,10 +301,7 @@ contract FPMMAdminTest is FPMMBaseTest {
     fpmm.configureTradingLimit(token0, config);
   }
 
-  function test_configureTradingLimit_whenOwner_shouldConfigureLimit()
-    public
-    initializeFPMM_withDecimalTokens(18, 18)
-  {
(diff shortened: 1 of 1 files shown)


contracts/swap/FPMM.sol
```diff
@@ -314,6 +319,17 @@ contract FPMM is IRPool, IFPMM, ReentrancyGuardUpgradeable, ERC20Upgradeable, Ow
     }
   }
 
+  /// @inheritdoc IFPMM
+  function getTradingLimits(
+    address token
+  ) external view returns (ITradingLimitsV2.Config memory config, ITradingLimitsV2.State memory state) {
+    FPMMStorage storage $ = _getFPMMStorage();
+    require(token == $.token0 || token == $.token1, "FPMM: INVALID_TOKEN");
```
nvtaveras: let's use custom errors in these to be consistent with the rest of the codebase. Maybe also in the library?

PR #670, opened 2026-01-05
fix: check factory on zap

bayological:
### Description

Validate factory address before zap in/out

### DIFF
--- a/contracts/swap/router/Router.sol
+++ b/contracts/swap/router/Router.sol
@@ -40,6 +40,10 @@ contract Router is IRouter, ERC2771Context {
     if (deadline < block.timestamp) revert Expired();
   }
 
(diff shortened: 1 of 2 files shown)


contracts/swap/router/Router.sol
```diff
@@ -40,6 +40,10 @@ contract Router is IRouter, ERC2771Context {
     if (deadline < block.timestamp) revert Expired();
   }
 
+  function _validateLiquidityFactory(address factory) internal view {
+    if (factory != defaultFactory) revert InvalidFactoryForLiquidity();
```
bowd: Why is only the defaultFactory considered a valid factory?

nvtaveras (changes requested):
see past comment

PR #674, opened 2026-01-14
fix: update reserves at the start of rebalance

### LINKED ISSUE #676: Rebalance Can Be DoS’d With 1-Wei Token Transfer
https://cantina.xyz/code/fe8e03b1-f8db-4698-8a21-44b3a862e434/findings/167

Mouradif:
**Finding:** https://cantina.0dns.co/224

### Description

During rebalance, the FPMM considers `amount0In` and `amount1In` as the diff between the balance and the reserve. If someone sends as little as 1 wei to the FPMM before the rebalance, then the rebalance will fail.

This PR adds a call to `_update()` at the start of the rebalance so that any received donations are incorporated into the reserve.

### Other changes

Added a test that fails if the issue isn't fixed.

### Related issues

- Fixes #676

### DIFF
--- a/test/integration/v3/CDPFPMM.DoS.t.sol
+++ b/test/integration/v3/CDPFPMM.DoS.t.sol
@@ -68,8 +68,6 @@ contract CDPFPMM_Test_SP_REBALANCE_FRONTRUN is
 
     $tokens.usdm.transfer(address($fpmm.fpmmCDP), 1);
     vm.prank(user);
-    // REBALANCE WILL REVERT WITH RebalanceDirectionInvalid()
-
     $liquidityStrategies.cdpLiquidityStrategy.rebalance(address($fpmm.fpmmCDP));
   }
 }


test/integration/v3/CDPFPMM.DoS.t.sol
```diff
@@ -0,0 +1,75 @@
...
+    $tokens.usdm.transfer(address($fpmm.fpmmCDP), 5_200e23);
+    $fpmm.fpmmCDP.swap(0, 10_000e23 - 1, user, "");
+    vm.stopPrank();
+    // REBALANCE WILL ATTEMPT TO TRANSFER EUR.M, AND TAKE OUT USD.M
+    // ATTACKER CAN FRONTRUN WITH A 1 WEI USD.M TRANSFER TO DOS THE REBALANCE
+    vm.prank(attacker);
+
+    $tokens.usdm.transfer(address($fpmm.fpmmCDP), 1);
+    vm.prank(user);
+    // REBALANCE WILL REVERT WITH RebalanceDirectionInvalid()
```
bowd: I'm guessing this comment was for when it reverted before the fix, but let's remove it or say rebalance succeeds or smth.

bowd (approved):

test/integration/v3/CDPFPMM.DoS.t.sol
Mouradif: Yep, that's from the researcher's PoC. Will fix now

PR #675, opened 2026-01-14
fix: optimize for contract size

Mouradif:
### Description

Change optimizer settings in `optimized` profile to make contracts fit the contract size limit (notably `TroveManager`)

### Doubts

Should this be the default settings? And we would leave `optimized` at 200 runs?

### DIFF
--- a/foundry.toml
+++ b/foundry.toml
@@ -24,7 +24,7 @@ verbosity = 3
 
 [profile.optimized]
 optimizer = true
-optimizer_runs = 200
+optimizer_runs = 1
 
 [profile.fork-tests]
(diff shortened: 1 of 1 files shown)


Mouradif:
@bowd This setting seems to apply to all compiled contracts including dependencies (bold).

Without this change:

```
# current work dir is this repo's root dir
$ FOUNDRY_PROFILE=optimized forge build --sizes

[...]

TroveManager       | 24,636   (60 bytes too many)
```

With this change:

```
$ FOUNDRY_PROFILE=optimized forge build --sizes

[...]

TroveManager       | 24,345   (we're in the clear)
```

But I'm closing this in favor of having these settings in the deployment repo

PR #688, opened 2026-01-22
docs: FPMM rounding analysis and protocol-favorable rounding recommendations

bowd:
## Summary

This PR adds a comprehensive analysis document covering all rounding points in FPMM pricing functions with recommendations for rounding in favor of the protocol.

### Document Contents

1. **Current Rounding Behavior** - Documents that all FPMM divisions currently truncate (round down)

2. **Rounding Points Analysis** - Detailed analysis of 8 key rounding locations:
   - Swap output calculation (`getAmountOut`)
   - Fee calculation in swap validation
   - Initial and subsequent liquidity minting
   - Liquidity burning
   - Price and price difference calculations
   - Rebalance incentive validation
   - Total value calculations

3. **Recommendations** - Priority changes identified:
   - **High:** Fee calculations should round UP
   - **High:** Rebalance minimum input should round UP
   - **Medium:** Price difference could round UP for conservative thresholds

4. **Impact on Rebalancing** - How changes would affect:
   - Amount conversion functions in `LiquidityStrategyTypes.sol`
   - Target amount calculations in `LiquidityStrategy.sol`
   - Incentive calculations
   - FPMM validation alignment
   - Decimal scaling functions

5. **Implementation Considerations** - Covers gas costs, consistency requirements, testing needs, and migration strategy

## Test plan
- [ ] Review document for technical accuracy
- [ ] Validate recommendations against actual FPMM behavior
- [ ] Use as reference for future rounding-related changes

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


mento-val (changes requested):
## QA Review — REQUEST CHANGES

**Verdict: Request Changes (one blocker — factual error in documentation)**

The rounding analysis document is well-reasoned, honest about its limitations, and the "Implementation Challenges Discovered" section is genuinely excellent institutional knowledge. I recommend merging once the single blocker below is resolved.

---

### Blocker: Fabricated comment in `_rebalanceCheck` analysis

**Section:** "Current Tolerance Mechanism" (~line 478 of `docs/FPMM_ROUNDING_ANALYSIS.md`)

The document presents this as live code:
```solidity
// allow for 10 wei difference due to rounding and precision loss
if (swapData.amount0In < minAmount0In) revert InsufficientAmount0In();
```

This comment **does not exist** in the actual `_rebalanceCheck` function. The real implementation is a strict check with no comment and no tolerance:
```solidity
if (swapData.amount0In < minAmount0In) revert InsufficientAmount0In();
```

If this comment was removed during the branch history, that should be stated explicitly — not presented as current behavior. Readers will look for this tolerance, not find it, and distrust the document. Worse, future implementers may assume tolerance exists where it does not.

**Fix:** Remove the fabricated comment from the snippet, or add a note: "Note: this tolerance comment was present during development but was removed; the current code is a strict check."

---

### Non-blocking suggestions

1. **`_transferProtocolFee` missing from rounding table** — this rounds DOWN (user-favorable), creating a small gap between "fee collected" and "fee sent to recipient". Worth a row for completeness.

2. **Overflow risk in `_convertWithRateAndFee`** — the formula multiplies four terms before dividing: `amount * numerator * toDecimals * incentiveNum`. With a 1e18-denominated oracle rate and meaningful amounts, intermediate products approach overflow territory. The doc recommends `Math.mulDiv` elsewhere but doesn't flag this existing formula has the same profile.

3. **`> 18 decimals` concern is unreachable** — the `getPrices` warning is technically correct but `initialize()` already reverts with `InvalidTokenDecimals` for tokens with > 18 decimals. Adjust the framing to reflect this guard.

4. **`_swapCheck` is stronger than described** — it enforces a total reserve-value invariant, not just fee arithmetic. This is a more important guarantee and worth stating clearly.

---

### Security assessment

The rounding design as a whole is sound: all value-transfer rounding truncates (dust stays in pool), and `_swapCheck`'s reserve-value invariant closes the door on rounding-based extraction attacks. The main open risk — tight coupling between rebalancing validation and FPMM math — is real and worth adding to NatDoc.

PR #719, opened 2026-03-12
feat: use msg.sender in clamp to not block determin action function

philbow61:
### Description

This Pr prevents the external determinAction function on the base class from being blocked by the introduction of the transient storage slot.  Before the clamp function, read the storage slot to get the balance of the rebalancer, but the slot was only set as part of the _beforeRebalance function. So when the external determineAction function was called, the slot was never set. Now we read the balance of msg.sender in the clamp functions, this is safe because there is no hook involved, so the msg.sender will be the same that called rebalance/determinAction. 
When the msg.sender is address(0), we don't clamp to get the ideal rebalancing amounts.


bayological (approved):

bayological (dismissed):
Nice work 👍

bowd (approved):

bayological (approved):
LGTM

PR #720, opened 2026-03-12
fix: add scoped remapping for bold library src imports

nvtaveras:
## Summary

This fixes the issue of the repo not building on neither the latest foundry stable version nor the nightly:

- The bold submodule uses `src/` relative imports internally (e.g. `import "src/Interfaces/IInterestRouter.sol"`)
- Without a scoped remapping, Foundry resolves these against the top-level project root and the build fails
- Adds `lib/bold/:src/=lib/bold/contracts/src/` to `remappings.txt` so those imports resolve correctly

## Test plan
- [x] `forge build` completes without errors

### DIFF
--- a/remappings.txt
+++ b/remappings.txt
@@ -1,4 +1,5 @@
 lib/bold/:openzeppelin-contracts/=lib/bold/contracts/lib/openzeppelin-contracts/
+lib/bold/:src/=lib/bold/contracts/src/
 openzeppelin/=lib/openzeppelin-contracts/
 openzeppelin-solidity/=lib/openzeppelin-contracts/
 openzeppelin-contracts/=lib/openzeppelin-contracts-next/


mento-val (approved):
The scoped remapping is correct and follows the same convention already used for the bold/openzeppelin import on the preceding line. Foundry scoped remapping syntax (`lib/bold/:src/=lib/bold/contracts/src/`) properly restricts the resolution to files under `lib/bold/`, pointing them at the correct `contracts/src/` path.

All CI passes (Lint & Test, Slither, Echidna). Minimal, targeted change — no security implications.

One small note: if the bold submodule also has other internal path prefixes beyond `src/` (e.g. `test/` or `script/`), those may need similar entries in the future — but that is not a blocker here.

chapati23 (approved):

philbow61 (approved):

